PR #1, opened 2021-08-13
Move CI files from libbpf

tsipa:
This also implements clone for CI helpers.

This changes has to be merged first.


anakryiko (commented):
While we are doing this, can we get rid of "travis" terminology and replace it with generic "ci"?

anakryiko (approved):
Given this blocks other things, I'm approving. But really consider getting rid of confusing travis mentions everywhere as a follow up.

tsipa:
I'd prefer step by step approach.
I will work on bash scripts separately as they concerns me a lot in general - as is it's somewhat hard to manage, but changing them or naming is not the intention of this change.

tsipa:
Only those with write access to this repository can merge pull requests.

PR #9, opened 2022-02-17
Workaround stdout issue

thefallentree:
also log qemu serial console output to boot.log for later use.


anakryiko:
probably should have squashed those 9 commits into one before merging?

thefallentree:
> probably should have squashed those 9 commits into one before merging?

it is squashed , I guess the github notification email didn't make it clear.

anakryiko:
I was looking at Github UI, it's still not clear even there. There is even "merged 9 commits" at the top. But yeah, seems like the final merged commit is just a single commit. Never mind.

Screen Shot 2022-02-17 at 10 49 31 AM

PR #20, opened 2022-07-21
Topic/more actions pr

danielocfb:
Instead of having a bunch of shell scripts here and a bunch more over in `kernel-patches/vmtest`, let's consolidate more here. Doing so will allow us to remove a bunch of duplicated scripts (such as `helpers.sh`, which I've counted to exist four times!).

This change adds actions that can be invoked by `vmtest` (or others), alongside the already existing ones performing similar tasks.

### DIFF
--- a/build-selftests/build_selftests.sh
+++ b/build-selftests/build_selftests.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+set -euo pipefail
+
+THISDIR="$(cd $(dirname $0) && pwd)"
+
+source "${THISDIR}"/../helpers.sh
+
+VMLINUX_BTF="$1"
+KERNEL="$2"
+TOOLCHAIN="$3"
+TOOLCHAIN_NAME="$(echo $TOOLCHAIN | cut -d '-' -f 1)"
+TOOLCHAIN_VERSION="$(echo $TOOLCHAIN | cut -d '-' -f 2)"
+
+if [ "$TOOLCHAIN_NAME" == "llvm" ]; then
+export LLVM="-$TOOLCHAIN_VERSION"
+LLVM_VER=$TOOLCHAIN_VERSION
+else
+LLVM_VER=15
+fi
+
+foldable start build_selftests "Building selftests with $TOOLCHAIN"
+
+LIBBPF_PATH="${REPO_ROOT}"
+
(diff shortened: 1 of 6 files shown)


build-selftests/build_selftests.sh
```diff
@@ -0,0 +1,52 @@
...
+set -euo pipefail
+
+THISDIR="$(cd $(dirname $0) && pwd)"
+
+source "${THISDIR}"/../helpers.sh
+
+VMLINUX_BTF="$1"
+KERNEL="$2"
+TOOLCHAIN="$3"
+TOOLCHAIN_NAME="$(echo $TOOLCHAIN | cut -d '-' -f 1)"
```
mykola-lysenko, line 12: can we move lines upto line 20 to some kind of a helper shell scrip that will determine tool chain? it seems these are repeated in all 3 shell files

mykola-lysenko:
Thanks Daniel for this clean-up! Really like where it all moving.

How did you test this?

danielocfb:
Thanks for the review.

> How did you test this?

I [hooked it up](https://github.com/danielocfb/vmtest/commits/topic/use-libbpf-ci-actions) in `vmtest` and then initiated a CI run with it. Right now these actions are not being used until we explicitly pull them in.

build-selftests/build_selftests.sh
danielocfb: I'd think that arguments to the script should almost certainly be assigned in the script itself and not some sourced helper, otherwise we most likely end up obfuscating what the script expects as input. A similar argument goes for options setting the behavior of the shell and the sourcing of the helper script, in my opinion. That should be done in the top level script itself, or we just end up with non-obvious state changes.

For the setting of `LLVM_VER` I tend to agree, but only insofar as that could be a function, not that the variable setting should happen in a different file. That being said, I'd think that should be done as part of a separate change. Except for the samples action, the scripts are just copied and pasted from the existing source and I don't think we should modify them in significant ways while we copy them around. So this really is an issue in existing code. What do you think?

mykola-lysenko: yeah, I was talking about TOOLCHAIN_NAME and VERSION parsing. we can do it as a separate change, fine with me. it is just we have it the same in all 3 files which makes me worried in case we update any of these

danielocfb: Sounds good. I will create a follow up change to factor this logic out.

PR #23, opened 2022-07-29
Revert "Rely on upstream kernel configuration"

danielocfb:
This change reverts commit 92f123b240be8b5fbebd08c7e5e2460b87e67e09. The
remaining instability in CI (caused by a NULL pointer dereference in the
kernel itself; reported [here]) seems to be caused by changes made as
part of minimizing the kernel configuration.
Let's rely on the non-upstream kernel configuration for now and until we
know which options precisely we should add (and/or fix the bug in
question).

[here]: https://lore.kernel.org/bpf/d8184db7-3247-c75f-7797-d3a5b0488743@huawei.com/T/#m1ea0cf0140edbd60d173f3ad52732db19a48cf2c

### DIFF
--- a/build-linux/build.sh
+++ b/build-linux/build.sh
@@ -16,8 +16,7 @@ fi
 
 foldable start build_kernel "Building kernel with $TOOLCHAIN"
 
-cat ${GITHUB_WORKSPACE}/tools/testing/selftests/bpf/config > .config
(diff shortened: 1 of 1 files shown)


danielocfb:
I had a test run including this change already [here](https://github.com/kernel-patches/bpf/actions/runs/2761315934) and it succeeded instantly.

mykola-lysenko (approved):

PR #24, opened 2022-08-01
Bump LLVM version we use to 16

danielocfb:
Development on LLVM 16 has started and version 15 is no longer available
in the repository we install it from. Bump the version we use
accordingly.


danielocfb:
Going to merge to get this in and CI back working (which will require additional changes in dependent repositories). If there are additional comments, I am happy to address them subsequently.

PR #26, opened 2022-08-08
Stop sourcing -latest kernel config

danielocfb:
https://github.com/kernel-patches/vmtest/commit/e5a73f5867617c873268914c229edbf8d407bf48
has renamed the kernel configuration in the kernel-patches/vmtest
repository, removing the '-latest' suffix.
As such, there is no need for us to keep attempting to source this file.
Stop doing so.


danielocfb:
Test runs are scheduled here:
https://github.com/kernel-patches/bpf/actions/runs/2819848666
https://github.com/kernel-patches/bpf/actions/runs/2819846582

danielocfb:
This change is really only a cleanup, there should be no functional difference, as the file got renamed in the repository.

PR #27, opened 2022-08-09
Introduce get-llvm-version action

danielocfb:
Whenever development on a new LLVM version starts, earlier versions of
llvm, clang, and lld may seize to exist in distribution snapshot
repositories. For us, that means having to manually bump the version in
three repositories [0][1][2], in a specific order. That's cumbersome to
do and ultimately we probably want a single source of truth as to what
the most recent LLVM version to use is.
This change introduces an action that centralizes this information while
making it available across repository boundaries. We will subsequently
be able to eliminate hardcoded version information in the other two
repositories constituting the BPF CI.

[0] https://github.com/kernel-patches/vmtest/pull/107
[1] https://github.com/libbpf/libbpf/pull/542
[2] https://github.com/libbpf/ci/pull/24

### DIFF
--- a/get-llvm-version/action.yml
+++ b/get-llvm-version/action.yml
@@ -1,0 +1,12 @@
+name: 'get-llvm-version'
+description: 'Retrieve the LLVM version to use'
+outputs:
+  version:
+    description: "LLVM version"
+    value: ${{ steps.llvm-version.outputs.version }}
+runs:
+  using: "composite"
+  steps:
+    - id: llvm-version
+      shell: bash
+      run: echo "::set-output name=version::$(source $GITHUB_ACTION_PATH/../helpers.sh; llvm_default_version)"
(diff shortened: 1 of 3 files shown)


danielocfb:
This pull request illustrates usage: https://github.com/kernel-patches/vmtest/pull/114

danielocfb:
Have you had a chance or do you intend to look at this change, @mykola-lysenko ?

danielocfb:
I think that in the future it *may* be possible to move away from an action that is actively shared between repositories to just having all logic pertaining LLVM versioning centralized here, but let's move on step at a time.

mykola-lysenko (approved):

mykola-lysenko:
Thanks @d-e-s-o for putting it together, I like that approach and direction.

PR #31, opened 2022-08-17
Make Qemu image path configurable

danielocfb:
The path to the file system image used by our invocation of Qemu is
currently hard coded to /tmp/root.img. This constant originates in the
run_vmtest.sh script, but is then hard-coded in other locations as well,
without an obvious connection between the two. That makes it hard to
understand the code.
This change makes the path to the image an explicit input parameter to
the prepare-rootfs action and the run_vmtest.sh script, which will make
it possible to make the input-output relationship more obvious in client
code moving forward.

### DIFF
--- a/prepare-rootfs/action.yml
+++ b/prepare-rootfs/action.yml
@@ -1,6 +1,6 @@
 name: 'prepare-rootfs'
 description: 'build rootfs
-1. download base root img to /tmp/root.img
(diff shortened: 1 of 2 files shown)


danielocfb:
Note that right now the input argument has a default value in order to not break any of the clients of this action. Once they are adjusted we can (and should) remove the default.

mykola-lysenko (approved):

PR #32, opened 2022-08-19
Plumb through test to run

danielocfb:
We would like to have the ability to run selftests in parallel as
opposed to fully serial as is currently the case. In order to make that
possible, this change adjusts the `prepare-rootfs` action to accept an
optional argument specifying the test to run and passes it to the
`run_selftests.sh` script that we adjusted already to earlier.


danielocfb:
I had scheduled a test run of `libbpf` with this change here: https://github.com/libbpf/libbpf/pull/555

mykola-lysenko (approved):

PR #34, opened 2022-08-29
Remove default arguments from prepare-rootfs action

danielocfb:
Having default arguments on an action should probably be avoided if
possible. For one, it's confusing to see something being "required" but
having a default value, but if not explicitly mentioned on the call
site, it's also unnecessarily hard to understand what is going on.

With https://github.com/libbpf/libbpf/pull/570 merged we no longer need
default values for two arguments to the prepare-rootfs action, as all
call sites explicitly provide values. Hence, remove them.

### DIFF
--- a/prepare-rootfs/action.yml
+++ b/prepare-rootfs/action.yml
@@ -14,12 +14,10 @@ inputs:
     description: 'arch'
     required: true
   kernel:
-    description: 'kernel version'
-    default: 'LATEST'
+    description: 'kernel version; use LATEST to use the most recent version'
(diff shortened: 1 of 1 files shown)


danielocfb:
vmtest run: https://github.com/kernel-patches/bpf/actions/runs/2951487973

chantra:
Thanks! Looking forward for this. Following github actions has been very confusing in the past and I agree that being explicit here is probably better.

mykola-lysenko (approved):

PR #35, opened 2022-08-29
Uppercase vmlinuz variable

danielocfb:
We typically used fully uppercased names for environment variables in
shell scripts. There does not seem to be a good reason for diverging for
$vmlinuz specifically. Rename it.


chantra:
LGTM

PR #36, opened 2022-08-29
Add kernel-root argument to run-qemu action

danielocfb:
This change adds the kernel-root argument to the run-qemu action. We
will need this argument once we want to move bpftool checks into the
actual selftest run job, as opposed to having them happen in the
unrelated prepare-rootfs. We need a default value while all "callers"
are being transitioned to providing this argument. The KERNEL_ROOT
variable that we set based on it is not currently unused (will come with
subsequent changes).

### DIFF
--- a/run-qemu/action.yml
+++ b/run-qemu/action.yml
@@ -10,6 +10,9 @@ inputs:
   vmlinuz:
     description: 'vmlinuz path'
     required: true
+  kernel-root:
+    description: 'kernel source dir'
+    default: '.'
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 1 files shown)


chantra:
What is consuming KERNEL_ROOT environment variable? Would it make sense to have the related diffs part of the same PR to better understand how everything fits together?

danielocfb:
> What is consuming KERNEL_ROOT environment variable? Would it make sense to have the related diffs part of the same PR to better understand how everything fits together?

Yeah, it would; but it's not easily possible because of inter-repository dependencies, unfortunately. Haven't found a satisfactory way to work around that.

It will be necessary in order for us to move the bpftool tests from the prepare-rootfs into run-qemu.

danielocfb:
This is the next change: https://github.com/danielocfb/libbpf-ci/commit/d6cbd163ebbe6d7a1411ffc2a27de30da64478c4

Long term we should move it into a single job (right now it runs as part of every selftest), but this is strictly better than what we currently have, which has been shown to be confusing for people debugging bpftool test failures.
We should probably also untangle the `KERNEL_ROOT`, `REPO_ROOT`, `REPO_PATH`, and `PROJECT_NAME` zoo of variables, but that's future work as well. Let's take one step at a time.

mykola-lysenko (approved):

PR #37, opened 2022-08-30
Move bpftool check into run-qemu action

danielocfb:
Currently the bpftool tests are run as part of the prepare-rootfs
action. That's counter-intuitive, as running a test has nothing
"preparational" to it. That in turn causes confusion with users that are
trying to debug failures of said tests: they have no idea how to find
the logs in question, because success/failure is reported as part of a
later action.
This change moves the running of the bpftool tests into the run-qemu
action, which is where selftest results are reported as well. As such,
we remove the need for an implicit file with a hard coded path being
used for communication between these two steps, and of course the
confusion for users is fixed as well.
It is still sub-optimal because we are running this test for every other
selftest run, but that is unchanged from before. The remaining issues
will be addressed at a later point.

### DIFF
--- a/run-qemu/run.sh
+++ b/run-qemu/run.sh
@@ -5,6 +5,25 @@ trap 'exit 2' ERR
 
 source $(cd $(dirname $0) && pwd)/../helpers.sh
 
+foldable start bpftool_checks "Running bpftool checks..."
+bpftool_exitstatus=0
+if [[ "${KERNEL}" = 'LATEST' ]]; then
+	# "&& true" does not change the return code (it is not executed if the
+	# Python script fails), but it prevents the trap on ERR set at the top
+	# of this file to trigger on failure.
+	"${REPO_ROOT}/${KERNEL_ROOT}/tools/testing/selftests/bpf/test_bpftool_synctypes.py" && true
+	bpftool_exitstatus=$?
+	if [[ $bpftool_exitstatus -eq 0 ]]; then
+		echo "bpftool checks passed successfully."
+	else
+		echo "bpftool checks returned ${bpftool_exitstatus}."
+	fi
+else
+	echo "bpftool checks skipped."
+fi
+bpftool_exitstatus="bpftool:${bpftool_exitstatus}"
+foldable end bpftool_checks
+
 foldable start vm_init "Starting virtual machine..."
 
 echo "Starting VM with $(nproc) CPUs..."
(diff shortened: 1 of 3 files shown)


danielocfb:
Please do not merge. Will take care of it once requisite changes landed.

danielocfb:
vmtest run: https://github.com/kernel-patches/bpf/actions/runs/2959736856

run-qemu/run.sh
```diff
@@ -5,6 +5,26 @@ trap 'exit 2' ERR
 
 source $(cd $(dirname $0) && pwd)/../helpers.sh
 
+foldable start bpftool_checks "Running bpftool checks..."
+bpftool_exitstatus=
+if [[ "${KERNEL}" = 'LATEST' ]]; then
+	# "&& true" does not change the return code (it is not executed if the
+	# Python script fails), but it prevents the trap on ERR set at the top
+	# of this file to trigger on failure.
+	"${REPO_ROOT}/${KERNEL_ROOT}/tools/testing/selftests/bpf/test_bpftool_synctypes.py" && true
```
mykola-lysenko, line 14: it is not happy with this variable not being set in https://github.com/libbpf/libbpf/runs/8102670715?check_suite_focus=true

danielocfb: I think it's fixed now: https://github.com/danielocfb/libbpf/runs/8102983329

mykola-lysenko: it failed s390x run for an unrelated reason, I retried and will finish review today

run-qemu/run.sh
```diff
@@ -5,6 +5,26 @@ trap 'exit 2' ERR
...
+	bpftool_exitstatus=$?
+	if [[ "$bpftool_exitstatus" -eq 0 ]]; then
+		echo "bpftool checks passed successfully."
+	else
+		echo "bpftool checks returned ${bpftool_exitstatus}."
+	fi
+	bpftool_exitstatus="bpftool:${bpftool_exitstatus}"
+else
+	echo "bpftool checks skipped."
```
mykola-lysenko, line 23: adding a comment on why such check is not needed for non 'LATEST' case would be beneficial

mykola-lysenko (approved):

chantra (commented):
Probably not worth doing in this specific commit, but could be in a commit on top.

The handling of  `bpftool_exitstatus` is convoluted. It seems we could do something line:

```
bpftool_exitstatus=0
if ...
...
else
...
fi
bpftool_exitstatus="bpftool:${bpftool_exitstatus}"
foldable end bpftool_checks
```

and get rid of all this  special handling of `bpftool_exitstatus=`

run-qemu/run.sh
danielocfb: I agree. But I don't know why it isn't, to be honest :-) If you know let me know and I can add it.

danielocfb:
> The handling of `bpftool_exitstatus` is convoluted.

Agreed. Adopted your suggestion.

danielocfb:
New set of CI runs:
- https://github.com/kernel-patches/bpf/actions/runs/2974138964
- https://github.com/libbpf/libbpf/actions/runs/2974141790

run-qemu/run.sh
qmonnet: This test is to check that everything is consistent between the commands supported by bpftool, its known type names, its documentation, its bash completion. We want to keep that in sync in the kernel, but there's no point in checking it for older kernels. If it failed on older kernels, it might be that some item was missing from the docs, not that some feature was broken.

So we want to run it on LATEST to make sure that nobody introduces a new type without updating bpftool everywhere necessary, but there's no interest in running this script for older kernels, we won't detect bugs or breakages.

PR #40, opened 2022-10-06
Add switch case to be able to run arm64 images in qemu

chantra:
The qemu parameters where tested as part of https://github.com/kernel-patches/rcu-rc/actions/runs/3193882128/jobs/5213462935

The `kvm_accel` values are very much a guess. This was only tested on AWS graviton host where kvm is not available.

### DIFF
--- a/run-qemu/run.sh
+++ b/run-qemu/run.sh
@@ -49,6 +49,14 @@ x86_64)
 	kvm_accel="-cpu kvm64 -enable-kvm"
 	tcg_accel="-cpu qemu64 -machine accel=tcg"
 	;;
+aarch64)
+	qemu="qemu-system-aarch64"
+	console="ttyAMA0,115200"
+	smp=$(nproc)
+	kvm_accel="-cpu host -enable-kvm -machine virt,gic-version=3,accel=kvm:tcg"
+	tcg_accel="-cpu cortex-a72 -machine virt,accel=tcg"
+	;;
+
 *)
 	echo "Unsupported architecture"
 	exit 1


run-qemu/run.sh
```diff
@@ -49,6 +49,14 @@ x86_64)
 	kvm_accel="-cpu kvm64 -enable-kvm"
 	tcg_accel="-cpu qemu64 -machine accel=tcg"
 	;;
+aarch64)
+	qemu="qemu-system-aarch64"
+	console="ttyAMA0,115200"
+	smp=$(nproc)
+	kvm_accel="-cpu max -machine virt,accel=kvm:tcg"
+	tcg_accel="-cpu cortex-a72 -machine virt,accel=kvm:tcg"
```
anakryiko, line 57: have you tried `-cpu host` instead of specifying exact CPU model? This would let qemu do faster emulation as it won't have to emulate different CPU instruction set?

chantra: I remember trying `-cpu max` fix mixed success. I can give this another spin. I was under the impression that because kvm is not available, it would just have to do total emulation anyway?

chantra:
I have updated the parameters after testing on a physical host too.

danielocfb (approved):
Not a `qemu`/`kvm` expert, but the change seems sensible to me.

chantra:
> Not a `qemu`/`kvm` expert, but the change seems sensible to me.

Neither am I :D 
But it did work in the few scenario I tried. It will probably evolve over time though.

PR #43, opened 2022-10-11
ci: Update images for all architectures

chantra:
Those images are all debian based and have an updated init script that symlinks /dev/fd to /proc/self/fd per https://github.com/libbpf/ci/pull/44

Signed-off-by: Manu Bretelle <chantr4@gmail.com>

PR #44, opened 2022-10-11
Support process substitution in vm

chantra:


chantra:
This images generated by those changes are being tested as part of the image bump in https://github.com/libbpf/ci/pull/43
https://github.com/kernel-patches/bpf/actions/runs/3223930145/jobs/5274523887

rootfs/mkrootfs_tweak.sh
```diff
@@ -33,6 +33,9 @@ fi
 /bin/mount bpffs /sys/fs/bpf -t bpf
 /bin/mount debugfs /sys/kernel/debug -t debugfs
 
+# Symlink /dev/fd to /proc/self/fd so process substitution work.
```
danielocfb, line 36: *works

rootfs/mkrootfs_debian.sh
```diff
@@ -15,6 +17,45 @@ if [ "$(id -u)" != 0 ]; then
 	exit 1
 fi
 
+function usage() {
+    echo "Usage: $0 [-a | --arch architecture] [-h | --help]
+
+Build a Debian chroot filesystem image for testing libbbpf in a virtual machine.
+By default build an image for the architecture of the host running the script.
```
danielocfb, line 24: It seems contradictory to say that the default arch is that of the host and then specifying that it's `amd64`.

danielocfb (approved):
Looks good to me.

rootfs/mkrootfs_debian.sh
chantra: ha, good point. I originally defaulted to `uname -m`, unfortunately, Debian uses amd64, arm64... instead of x86_64, aarch64 so I ended up defaulting to amd64.

Anywhay... that got me to dig a bit more as to how debootstrap would figure out which architecture to use in Debian lingua... this got me to find
```
$ /usr/bin/dpkg --print-architecture
amd64
```

PR #52, opened 2022-10-13
Downgrade rootfs from Bookworm to Bullseye

chantra:


chantra:
CI run in https://github.com/kernel-patches/bpf/actions/runs/3245852937/jobs/5323907665

PR #56, opened 2022-10-24
rootfs: fix foreign arch

### LINKED ISSUE #55: mkrootfs_debian.sh failing to build foreign architecture
Building x-arch rootfs started failing. This is because a change may have happen in bullseye recently that calls a binary as part of debootstrap:

The exact error is:
```
I: Extracting zlib1g...
W: Failure trying to run: chroot "/home/chantra/devel/libbpf/ci/rootfs/tmp.jwLB9EVVaG" /bin/true
W: See /home/chantra/devel/libbpf/ci/rootfs/tmp.jwLB9EVVaG/debootstrap/debootstrap.log for details
```
and the actual log line:
```
chroot: failed to run command '/bin/true': Exec format error
```

The solution to this is to use the `--foreign` option of debootstrap and using qemu.

Example: https://askubuntu.com/questions/287789/what-is-debootstrap-second-stage-for

chantra:
An update in upstream cause a binary on the foreign platform to be
executed, which in turned caused the rootfs creation to fail with
something like:
```
chroot: failed to run command '/bin/true': Exec format error
```
 
`debootstrap` can handle that natively by building the image in 2 steps.
1) download the packages, verify, and extract them inside the chroot
2) running a second stage that unpack and configure them.

Before the second stage, we copy `qemu-${foreign_arch}-static` into the
chroot, allowing the native binaries to run properly.

After this fix, the rootfs images could be built.

Fixed #55

### DIFF
--- a/rootfs/mkrootfs_debian.sh
+++ b/rootfs/mkrootfs_debian.sh
@@ -9,6 +9,7 @@
 # Any arguments that need to be passed to `debootstrap` should be passed after
 # `--`, e.g ./mkrootfs_debian.sh --arch=s390x -- --foo=bar
 
+set -e -u -o pipefail
 
 # table of Debian arch <-> GNU arch matches
 CPUTABLE="${CPUTABLE:-/usr/share/dpkg/cputable}"
@@ -48,7 +49,7 @@ function qemu_static() {
 
 function check_requirements() {
     # Checks that all necessary packages are installed on the system.
-    # print an error message explaining what is missing and exits.
+    # Prints an error message explaining what is missing and exits.
 
     local deb_arch=$1
     local err=0
@@ -69,14 +70,14 @@ function check_requirements() {
 
     # Check that we can install the root image for a foreign arch.
     qemu=$(qemu_static "${deb_arch}")
-    if [[ -z $(which "${qemu}") ]]
+    if ! command -v "${qemu}" &> /dev/null
     then
         error "${qemu} binary not found on your system. Make sure qemu-user-static package is installed."
         err=1
     fi
 
     # Check that debootrap is installed.
-    if [[ -z "$(which debootstrap 2> /dev/null)" ]]
+    if ! command -v debootstrap &> /dev/null
     then
         error "debootstrap binary not found on your system. Make sure debootstrap package is installed."
         err=1
@@ -129,7 +130,8 @@ done
 
 check_requirements "${deb_arch}"
 
-set -e -u -x -o pipefail
+# Print out commands ran to make it easier to troubleshoot breakages.
+set -x
 
 # Create a working directory and schedule its deletion.
 root=$(mktemp -d -p "$PWD")
@@ -151,7 +153,22 @@ packages=(
 	zlib1g
 )
 packages=$(IFS=, && echo "${packages[*]}")
-debootstrap --include="$packages" --variant=minbase --arch="${deb_arch}" "$@" "${distro}" "$root"
+
+# Stage 1
+debootstrap --include="$packages" \
+    --foreign \
+    --variant=minbase \
+    --arch="${deb_arch}" \
+    "$@" \
+    "${distro}" \
+    "$root"
+
+qemu=$(which $(qemu_static ${deb_arch}))
+
+cp "${qemu}" "${root}/usr/bin"
+
+# Stage 2
+chroot "${root}" /debootstrap/debootstrap --second-stage
 
 # Remove the init scripts (tests use their own). Also remove various
 # unnecessary files in order to save space.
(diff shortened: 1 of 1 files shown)


rootfs/mkrootfs_debian.sh
```diff
@@ -26,11 +22,78 @@ function usage() {
...
+    # Check that the architecture is supported  by Debian.
+    if [[ -z $(debian_to_gnu "${deb_arch}") ]]
+    then
+        error "${deb_arch} is not a supported architecture."
+        err=1
+    fi
+
+    # Check that we can install the root image for a foreign arch.
+    qemu=$(qemu_static "${deb_arch}")
+    if [[ -z $(which "${qemu}") ]]
```
danielocfb, line 72: I believe that `command -v` is the typical way for checking whether a command exists (https://manpages.org/command) [not that it will make a big difference]

rootfs/mkrootfs_debian.sh
```diff
@@ -26,11 +22,78 @@ function usage() {
...
+function qemu_static() {
+    # Given a Debian architecture find the location of the matching
+    # qemu-${gnu_arch}-static binary.
+    gnu_arch=$(debian_to_gnu "${1}")
+    echo "qemu-${gnu_arch}-static"
+}
+
+function check_requirements() {
+    # Checks that all necessary packages are installed on the system.
+    # print an error message explaining what is missing and exits.
```
danielocfb, line 51: `# Prints an error message explaining what is missing and exits.`

rootfs/mkrootfs_debian.sh
```diff
@@ -63,6 +126,11 @@ while true; do
     esac
 done
 
+
+check_requirements "${deb_arch}"
+
+set -e -u -x -o pipefail
```
danielocfb, line 132: May I ask why you moved this setting down? It seems to be most aptly placed at the top of the file from what I understand.

danielocfb (approved):
Looks mostly good to me. Left a few comments.

rootfs/mkrootfs_debian.sh
chantra: Fair. I moved it because otherwise it was overly verbose (due to -x). This made it harder to get any significant signal from the output.
```
[15:26:58] chantra@worktop:rootfs git:(rootfs_fix_foreign_arch) $ ./mkrootfs_debian.sh         
ERROR: ./mkrootfs_debian.sh must run as root
[15:27:05] chantra@worktop:rootfs git:(rootfs_fix_foreign_arch) $ vi mkrootfs_debian.sh 
[15:27:54] chantra@worktop:rootfs git:(rootfs_fix_foreign_arch*) $ ./mkrootfs_debian.sh 
+ CPUTABLE=/usr/share/dpkg/cputable
++ dpkg --print-architecture
+ deb_arch=amd64
+ distro=bullseye
++ getopt -l arch:,distro:,help -o a:d:h --
+ TEMP=' --'
+ '[' 0 -ne 0 ']'
+ eval set -- ' --'
++ set -- --
+ unset TEMP
+ true
+ case "$1" in
+ shift
+ break
+ check_requirements amd64
+ local deb_arch=amd64
+ local err=0
+ [[ ! -e /usr/share/dpkg/cputable ]]
++ debian_to_gnu amd64
++ awk -v deb_arch=amd64 '$1 ~ deb_arch {print $2}' /usr/share/dpkg/cputable
+ [[ -z x86_64 ]]
++ qemu_static amd64
+++ debian_to_gnu amd64
+++ awk -v deb_arch=amd64 '$1 ~ deb_arch {print $2}' /usr/share/dpkg/cputable
++ gnu_arch=x86_64
++ echo qemu-x86_64-static
+ qemu=qemu-x86_64-static
++ which qemu-x86_64-static
+ [[ -z /usr/bin/qemu-x86_64-static ]]
++ which debootstrap
+ [[ -z /usr/sbin/debootstrap ]]
++ id -u
+ [[ 1000 != 0 ]]
+ error './mkrootfs_debian.sh must run as root'
+ echo 'ERROR: ./mkrootfs_debian.sh must run as root'
ERROR: ./mkrootfs_debian.sh must run as root
+ err=1
+ [[ 1 -ne 0 ]]
+ exit 1
```

 I can probably move it back to the top (without -x) and only `set -x` here.

-z is only relevant when we start to do some relevant work so we can find out where the script break. Here this is just some basic validation happening.

rootfs/mkrootfs_debian.sh
```diff
@@ -151,15 +151,31 @@ packages=(
...
+# Stage 1
+debootstrap --include="$packages" \
+    --foreign \
+    --variant=minbase \
+    --arch="${deb_arch}" \
+    "$@" \
+    "${distro}" \
+    "$root"
+
+qemu_static=$(which $(qemu_static ${deb_arch}))
```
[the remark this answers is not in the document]
chantra: oh yeah, this was not intentional. good catch.

chantra:
addressed the comments.

Issue #54, opened 2022-10-23
init script spewing errors when vm starts


chantra:
The checks seem valid yes, when the vm starts we get this error in the logs (see https://github.com/kernel-patches/bpf/actions/runs/3303109637/jobs/5450787223#step:6:425)
```
+ [ -x /etc/rcS.d/S10-mount ]
  + /etc/rcS.d/S10-mount
  + /bin/mount proc /proc -t proc
  + /bin/mount -t devtmpfs
  + [[ -z devtmpfs on /dev type devtmpfs (rw,relatime,size=3029136k,nr_inodes=757284,mode=755) ]]
  [[: on: unknown operand
  + /bin/mount sysfs /sys -t sysfs
  + /bin/mount bpffs /sys/fs/bpf -t bpf
  + /bin/mount debugfs /sys/kernel/debug -t debugfs
  + [[ -a /dev/fd ]]
  [[: /dev/fd: unknown operand
  + ln -s /proc/self/fd /dev/fd
  + echo Listing currently mounted file systems
  Listing currently mounted file systems
  + /bin/mount
  /dev/vda on / type ext4 (rw,relatime)
```

chantra:
it is a busybox issue. 

```
root@(none) /]# busybox sh    
/ # [[ -z $(/bin/mount -t devtmpfs) ]]
sh: on: unknown operand
/ # [[ -a /dev/fd ]] 
sh: /dev/fd: unknown operand
```

sh:
```
[root@(none) /]# sh
sh-5.1# [[ -z $(/bin/mount -t devtmpfs) ]]
sh-5.1# [[ -a /dev/fd ]] 
```

It seems the problem is that if the shell invoking out init script is busybox, /bin/sh will be the one from busybox, which has the limitation.

```
[root@(none) /]# readlink "/proc/$$/exe"
/usr/bin/bash
[root@(none) /]# busybox sh
/ # readlink "/proc/$$/exe"
/usr/bin/busybox
/ # sh
/ # readlink "/proc/$$/exe"
/usr/bin/busybox
/ # bash
[root@(none) /]# readlink "/proc/$$/exe"
/usr/bin/bash
```

chantra:
solution is to wrap the command substitution in quotes, and replace -a with -e...

chantra:
so.... I have a solution for this, but this is currently blocked on #55

PR #58, opened 2022-10-24
rootfs: Fix init script errors #54

chantra:
Those images were built with both #56 and #57 applied.

The former fixes building the rootfs for a foregin architecture and should not result in any material changes in the image.

The latter fixes an initscript to work when run with busybox'sh.


chantra:
test run: https://github.com/kernel-patches/vmtest/actions/runs/3308807044/jobs/5462256604#step:6:425

danielocfb (approved):

PR #59, opened 2022-10-25
add shell linter

teknoraver:
This adds a GitHub action which runs a shell linter on all the .sh files.
ludeeus/action-shellcheck is just an action which relies on
https://github.com/koalaman/shellcheck which is a well known linter.

```bash
~/src/libbpf-ci$ find -type f -name '*.sh' |xargs shellcheck -S error
~/src/libbpf-ci$ echo $?
0
```

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,19 @@
+name: "lint"
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  shellcheck:
+    name: ShellCheck
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Run ShellCheck
+        uses: ludeeus/action-shellcheck@master
+        env:
+          SHELLCHECK_OPTS: -S error
--- a/helpers.sh
+++ b/helpers.sh
@@ -1,3 +1,5 @@
+# shellcheck shell=bash
+
 # $1 - start or end
 # $2 - fold identifier, no spaces
 # $3 - fold section description
--- a/prepare-rootfs/run.sh
+++ b/prepare-rootfs/run.sh
@@ -438,7 +438,7 @@ chmod 644 /exitstatus
 HERE
 
 # Create the init scripts.
-if [[ ! -z SETUPCMD ]]; then
+if [[ -n $SETUPCMD ]]; then
 	# Unescape whitespace characters.
 	setup_cmd=$(sed 's/\(\\\)\([[:space:]]\)/\2/g' <<< "${SETUPCMD}")
 	kernel="${KERNELRELEASE}"


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,17 @@
+name: "lint"
+
+on:
+  pull_request: {}
```
danielocfb, line 4: Is that the same as `pull_request:` (no curly brackets)? Just use that?

danielocfb (changes requested):
The code seems fine to me. Thanks!

Would you mind fixing the typo in the subject ("liter") and write a bit more of a commit message? E.g., why did you choose to use `ludeeus/action-shellcheck` and not any other variant?

Lastly, given that you created a pull request I assume you verified that the check works and reported no issues. Can you please provide a pointer to a sample run as well?

teknoraver:
oops, my bad, I'll fix the subject

danielocfb:
> > Would you mind fixing the typo in the subject ("liter") and write a bit more of a commit message? E.g., why did you choose to use `ludeeus/action-shellcheck` and not any other variant?
> 
> I guess that this is the most used one. Do you suggest another one?

I am saying that the reasoning is something that most likely should be contained in the commit message, so that in the future somebody can understand why a decision was made.

> > Lastly, given that you created a pull request I assume you verified that the check works and reported no issues. Can you please provide a pointer to a sample run as well?
> 
> The check works, but finds tons of warnings. I'm addressing all of them in my private repo, so to avoid spamming this one.

Either is fine, thanks!

teknoraver:
This is a sample run of the action on my fork:
https://github.com/teknoraver/libbpf-ci/actions/runs/3329735501/jobs/5507322095

rootfs/mkrootfs_tweak.sh
```diff
@@ -16,7 +16,7 @@ cat > "$root/etc/inittab" << "EOF"
 EOF
 chmod 644 "$root/etc/inittab"
 
-mkdir -m 755 -p "$root/etc/init.d" "$root/etc/rcS.d"
```
danielocfb, line 19: May I ask what the rationale for this change is? Doesn't seem semantic-preserving to me.

rootfs/mkrootfs_debian.sh
```diff
@@ -152,18 +152,18 @@ packages=(
 	strace
 	zlib1g
 )
-packages=$(IFS=, && echo "${packages[*]}")
```
danielocfb, line 155: NIt: In my opinion this rename is confusing. What is the complaint and can we disable it instead?

prepare-rootfs/run.sh
```diff
@@ -63,14 +64,12 @@ Miscellaneous:
...
   -h, --help           display this help message and exit"
-
+EOF
 	case "$1" in
 		out)
-			echo "$USAGE_STRING"
 			exit 0
 			;;
 		err)
-			echo "$USAGE_STRING" >&2
```
danielocfb, line 73: Okay, so earlier we were emitting the string to stderr, now we are no longer? If the title of the diff is "suppress warnings" then I infer that semantics are preserved, no? Unsure how to reconcile these two statements, to be honest. Can you please keep diffs focused on a single purpose and *explain* why something is done. I will not be the only one reading this code down the line and posing similar questions, so it would really help if we explained changes being made.

helpers.sh
```diff
@@ -44,9 +44,10 @@ llvm_default_version() {
 
 # $1 - toolchain name
 llvm_version() {
-  local toolchain="$1"
-  local toolchain_name="$(echo $toolchain | cut -d '-' -f 1)"
-  local toolchain_version="$(echo $toolchain | cut -d '-' -f 2)"
+  local toolchain, toolchain_name, toolchain_version
+  toolchain="$1"
+  toolchain_name="$(echo "$toolchain" | cut -d '-' -f 1)"
+  toolchain_version="$(echo "$toolchain" | cut -d '-' -f 2)"
```
danielocfb, line 50: Nitpicking here, but this is standard shell variable initialization to the degree I can tell and the new version is much more foreign to 99% of people, I'll argue. We should probably just disable this warning if it complains about *that*.

get-linux-source/checkout_latest_kernel.sh
```diff
@@ -26,27 +24,27 @@ if [ ! -d "${REPO_PATH}" ]; then
...
 	foldable start pull_kernel_srcs "Fetching kernel sources"
 
-	mkdir -p $(dirname "${REPO_PATH}")
-	cd $(dirname "${REPO_PATH}")
+	mkdir -p "$(dirname "${REPO_PATH}")"
+	cd "$(dirname "${REPO_PATH}")"
 	# attempt to fetch desired bpf-next repo snapshot
 	if [ -n "${SNAPSHOT_URL}" ] && wget -nv ${SNAPSHOT_URL} && tar xf bpf-next-${LINUX_SHA}.tar.gz --totals ; then
-		mv bpf-next-${LINUX_SHA} $(basename ${REPO_PATH})
+		mv "bpf-next-${LINUX_SHA}" "$(basename ${REPO_PATH})"
```
danielocfb: For my understanding, why is `REPO_PATH` not quoted here similar to how `$0` is at the top of the file?

download-vmlinux/run.sh
```diff
@@ -3,22 +3,15 @@
 set -uo pipefail
 trap 'exit 2' ERR
 
-source $(cd $(dirname $0) && pwd)/../helpers.sh
+. "$(cd "$(dirname "$0")" && pwd)/../helpers.sh"
 
-TEMP=$(getopt -o 'k:b:r:fos:ISid:lh' --long 'kernel:,build:,rootfs:,force,one-shot,setup-cmd,skip-image,skip-source:,interactive,dir:,list,help' -n "$0" -- "$@")
```
danielocfb, line 8: As discussed offline, let's get a CI run scheduled to get this tested.

And, actually, it will likely need to happen in the [`libbpf/libbpf`](https://github.com/libbpf/libbpf) repository, I believe, because only it downloads a tarball (and uses this action).

For my understanding, though, is this removal based on your detection of dead code or based on being flagged by `shellcheck`?

patch-kernel/patch_kernel.sh
```diff
@@ -3,8 +3,8 @@
 set -eu
 DIFF_DIR=$1
 
-if ls ${DIFF_DIR}/*.diff 1>/dev/null 2>&1; then
-  for file in ${DIFF_DIR}/*.diff; do
+if ls "$DIFF_DIR"/*.diff >/dev/null 2>&1; then
+  for file in "$DIFF_DIR"/*.diff; do
```
danielocfb, line 7: May I ask what is the rationale for removing the curly braces here but not elsewhere? It strikes me as random, to be honest.

danielocfb (commented):
Thanks for the update. Few comments.

> Disable style suggestions SC1091, SC2181 and SC2001 because they are not
> worth fixing in this repository.

Can you please explain what they represent?

rootfs/mkrootfs_debian.sh
danielocfb: Sorry, the highlight is weird. I am referring to the new name, `packages2`, not the old name.

rootfs/mkrootfs_tweak.sh
teknoraver: There is a warning saying that the attributes are only applied to the last path created.
Given that `$root/etc/` already exist, suppress that warning by removing `-p`

prepare-rootfs/run.sh
teknoraver: Yes, this was a mistake.
The complaint here is that the string was not terminated. Ideally we need to put a `\` at every end of line, which will clutter the file.
I can restore the stderr output with a simple change

get-linux-source/checkout_latest_kernel.sh
teknoraver: That can be quoted too. Anyway I can suggest a change which will get rid of all $(basename ...) stuff:
```
$ PATH=/var/lib/something
$ echo ${PATH%/*}        
/var/lib
```

download-vmlinux/run.sh
teknoraver: It was flagged by shellcheck

patch-kernel/patch_kernel.sh
teknoraver: That's not really needed. I need to remove the `1` which is redundant, and while at it I removed the {}

helpers.sh
teknoraver: That was the warning:
https://www.shellcheck.net/wiki/SC2155

teknoraver:
I refreshed a PR with a smaller one which only fixes the severe errors.
We'll take care of the many warnings later, along with other improvements.

danielocfb:
Here is a `vmtest` CI run: https://github.com/kernel-patches/bpf/actions/runs/3339312442
Here `libbpf`: https://github.com/danielocfb/libbpf/actions/runs/3339319849

danielocfb:
The pull request causes a new deprecation warning to show up: https://github.com/libbpf/ci/actions/runs/3339058943. We just removed those: https://github.com/kernel-patches/vmtest/issues/142. It's not blocking, but would you mind trying to get the warning removed, @teknoraver ?

teknoraver:
BTW according to GitHub, this deprecation was added last week, and deprecated commands will be available until 31st May 2023:

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Let's hope that the action will be fixed in the meantime!

danielocfb:
CI doesn't seem to have any objections. Neither do I. Thanks!

PR #63, opened 2022-11-02
ci: move s390x self-hosted builder into libbpf/ci

chantra:


chantra:
This was tested on current s390x runners. The second diff of this stack now shows in the label list
image

mykola-lysenko (approved):

PR #64, opened 2022-11-02
Ansible s390x import

chantra:
Import the existing ansible playbook/roles that are used to provision s390x and prepare ground to support other deployments by moving the tasks into its own file (instead of main.yml)


ansible/roles/runner/tasks/main.yml
```diff
@@ -1,209 +1,7 @@
...
-    - reload systemd daemon
-    - restart actions-runner-watchdog timer
-
-- name: Enable actions runner watchdog timer
-  become: yes
-  ansible.builtin.service:
-    name: actions-runner-watchdog.timer
-    enabled: yes
-    state: started
+- include_tasks: main-generic.yml
```
danielocfb, line 13: For my understanding, the non-existence of this file is fine because we don't use Ansible for anything but the `s390x` architecture currently, is that correct?

danielocfb (approved):
Not an expert in the configuration, but it makes sense to the degree I can tell.

ansible/roles/runner/tasks/main.yml
chantra: correct and good catch. I did add it later in a stack to come..... https://github.com/chantra/ci/commit/f2a1c2d96511958b463c1299f55f780bca9dda96

I wanted to get this base sorted before building on top of it (and was actively iterating on it).
The current final aws + s390x work is https://github.com/chantra/ci/commits/ansible_aws but I am working on a separate stack to have better indiviual commits.

PR #68, opened 2022-11-03
Honor patches with .patch extension

danielocfb:
This change prepares the `patch-kernel` action for the application of patches with the extension `.patch`, in addition to `.diff`. The need to rename patches to sport the `.diff` extension has repeatedly been shown to be a difficult thing to remember.
See https://github.com/kernel-patches/vmtest/issues/160 for additional technical details.


chantra (approved):
Thanks! This will make it easier to just `git format-patch` away.

PR #69, opened 2022-11-04
Ansible for AWS runners

chantra:
This stack does quite a few things:
- allow setting runners with ansible on AWS
-  use kvm when pssible (bare-metal) and also allow setting multiple runners per repo
- add architecture in the runner label see kernel-patches/vmtest#162 for context.
- add support to run runners using a GH app instead of a GH user, this helps runners not being impacted/impact rate limiting
- add a runner prefix

### DIFF
--- a/ansible/roles/runner/files/app_token.sh
+++ b/ansible/roles/runner/files/app_token.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+#
+# Request an ACCESS_TOKEN to be used by a GitHub APP
+# Environment variable that need to be set up:
+# * APP_ID, the GitHub's app ID
+# * APP_PRIVATE_KEY, the content of GitHub app's private key in PEM format.
+#
+# https://github.com/orgs/community/discussions/24743#discussioncomment-3245300
+#
+
+set -o pipefail
+
+_GITHUB_HOST=${GITHUB_HOST:="github.com"}
+
+# If URL is not github.com then use the enterprise api endpoint
+if [[ ${GITHUB_HOST} = "github.com" ]]; then
+  URI="https://api.${_GITHUB_HOST}"
+else
+  URI="https://${_GITHUB_HOST}/api/v3"
+fi
+
+API_VERSION=v3
+API_HEADER="Accept: application/vnd.github.${API_VERSION}+json"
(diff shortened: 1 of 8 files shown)


ansible/roles/runner/files/app_token.sh
```diff
@@ -0,0 +1,88 @@
...
+
+rs256_sign() {
+    openssl dgst -binary -sha256 -sign <(echo "$1")
+}
+
+request_access_token() {
+    jwt_payload=$(build_jwt_payload)
+    encoded_jwt_parts=$(base64url <<<"${JWT_JOSE_HEADER}").$(base64url <<<"${jwt_payload}")
+    encoded_mac=$(echo -n "$encoded_jwt_parts" | rs256_sign "${APP_PRIVATE_KEY}" | base64url)
+    generated_jwt="${encoded_jwt_parts}.${encoded_mac}"
```
danielocfb, line 70: Could these variables be `local`?

ansible/roles/runner/files/app_token.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
```
danielocfb, line 1: In my opinion, this kind of code should be written in Python which has more native support for HTTP requests, JSON, etc.

danielocfb (approved):
Looks good from what I can tell. Some superficial comments. Realizing that I reviewed in reverse order... :-(

ansible/roles/runner/files/app_token.sh
chantra: probably could, but I would rather keep this close to the version in https://github.com/myoung34/docker-github-actions-runner/pull/254 .

ansible/roles/runner/files/app_token.sh
chantra: This is based of the work I did in https://github.com/myoung34/docker-github-actions-runner/pull/254 , used bash because this is what upstream was using. 
I would rather not recreate a custom version for our use case. What I should probably do, now that it has been merged, is no keep the internal version here, put pull it directly from upstream, e.g from https://raw.githubusercontent.com/myoung34/docker-github-actions-runner/master/app_token.sh

PR #74, opened 2022-11-10
Allow passing the tests to run as an argument to qemu

chantra:
Historically, we build a VM for each individual tests we wanted to run. At the end of the day, all tests are available within the VM and it is just a matter of what we did put in our init script.
With kernel-patches/vmtest#139 we can now pass the test to run as a command line argument at boot time. So, the run-qemu step will need to provide that test name.

tested with an unmodified kernel-patches/vmtest: https://github.com/kernel-patches/vmtest/pull/171/checks
and a modified one (e.g building the rootfs vm after building the kernel and providing the test on the boot command):  https://github.com/kernel-patches/vmtest/actions/runs/3432974206/jobs/5723083313

### DIFF
--- a/run-qemu/action.yml
+++ b/run-qemu/action.yml
@@ -16,6 +16,9 @@ inputs:
   max-cpu:
     description: 'Maximum number of CPU allocated to a VM (regardless of number of CPUs available on the host). Default is unset, e.g it will default to the number of CPU on the host.'
     default: ''
+  kernel-test:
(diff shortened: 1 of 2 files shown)


danielocfb (approved):
Looks good to me.

chantra:
> Looks good to me.

Thanks for the quick turnover. I updated the  description to be more comprehensive.

PR #76, opened 2022-11-14
Make kbuild-output input required in all actions

danielocfb:
At this point, there is no reason to build without KBUILD_OUTPUT specified: maintaining additional code paths that build that way is just a burden and there is no benefit of building within the source tree versus outside of it.
Hence, this change proposes making the kbuild-output argument to the various actions mandatory. It's already provided by all clients, so there is no effective change.

### DIFF
--- a/build-linux/action.yml
+++ b/build-linux/action.yml
@@ -9,7 +9,7 @@ inputs:
     default: 'gcc'
   kbuild-output:
     description: 'relative or absolute path to use for storing build artifacts'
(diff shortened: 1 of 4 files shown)


danielocfb:
libbpf: https://github.com/libbpf/libbpf/actions/runs/3464283925
bpf: https://github.com/kernel-patches/bpf/actions/runs/3464298584

chantra (approved):

PR #79, opened 2022-11-17
Print diff when updating kernel config to use

danielocfb:
The `build-linux` action conditionally updates the kernel config to use if it changed. It's generally useful to be able to see what exactly has changed. For example, the (expanded) configuration contains the compiler version and given that we install the newest clang snapshot on each run we may end up with a new compiler version compared to the one that created the build artifacts and  that forces a full rebuild. It's generally non-obvious when that happens and so printing a diff of the two configs can help better understand behavior, should the need arise.


danielocfb:
Here is what the output may look like in practice: https://github.com/danielocfb/kernel-patches-bpf/actions/runs/3493964811/jobs/5849322937#step:10:46

mykola-lysenko (approved):

PR #80, opened 2022-11-28
ansible: use arch specific runners

chantra:
With s390x runners being built by kernel-patches/runner, we can now get rid of the logic to build the runners on s390x and instead pull the readily available s390x runner from the container repository.


danielocfb (approved):
Seems fine to me.

chantra:
This has been pushed on all servers. Landing.

PR #81, opened 2023-01-18
[ansible][ci] Disable runner auto-update

chantra:
When the containers try to auto-update, they need to restart, which cause them to come back on the old version. If the server instruct the runner to update again, we can end up in continuously restarting the runner without doing any meaningful work.

Our runners are being updated by a separate job [0] which produce a new container nightly and eventually pick up the update, so we don't really need the auto-update feature.

[0] https://github.com/kernel-patches/runner


chantra:
This was tested by pushing the change to our runners.

danielocfb (approved):
Looks reasonable to me, thanks.

PR #86, opened 2023-02-08
default to llvm 16

chantra:
This change get us back onto llvm-16 by default as it seemsthat currently, qemu VMs are failing to boot when the kernel is built with llvm 17, it is very likely related to
https://github.com/ClangBuiltLinux/linux/issues/1800 .

Because of the layout of LLVM's apt repo, unless the version is the latest they distribute, the repository distro need to use the version number as suffix.

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -39,6 +39,11 @@ print_notice() {
 
 # No arguments
 llvm_default_version() {
+  echo "16"
+}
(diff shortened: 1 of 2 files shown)


danielocfb (approved):
Looks good to me. Thanks for looking into this issue!

Issue #84, opened 2023-02-01
[prepare-rootfs] improve rootfs selection


chantra:
Currently, we run tests in a Debian Bullseye rootfs. This works with our current CI runners (Ubuntu focal) but this is very brittle.

1. in the past, we had some dependency issues.
2. people may run vmtests from different hosts then Ubuntu focal.

The way we do rootfs selection at the moment is pretty simple: https://github.com/libbpf/ci/blob/1b794627fd7431428a7e5a077010285733eaedf2/prepare-rootfs/run.sh#L223C26-L231

The rootfs have the distro release in their name, so we could possibly filter on this.
The script to generate rootfs uses debootstrap and should work with pretty much any Debian derivative.

We could generate a rootfs for common rootfs we care about, put them all in INDEX and when selecting the rootfs, try to match the hosts' distro, and fallback on a default one otherwise.

This should help with #83, and avoid issues like we had in the past (see comment in that PR).

Because currently libbpf/ci and vmtest in kernel tree differ, we would also need to update
https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/tree/tools/testing/selftests/bpf/vmtest.sh?id=c1a3daf7363b48c6c4b86aee2efa2287f695f649#n106  with the change.

Given the current behaviour of that function, we may be able to get away with updating the index file and the test will just pick up whatever is latest alphabetically.

iii-i:
I like this proposal, thanks for looking into it.
A small suggestion: it would also be nice to provide a way to specify the distro manually when running vmtest.sh.

chantra:
Since  #117 was landed and used in BPF CI, this should not be a problem anymore.

PR #87, opened 2023-02-13
Cross-compilation support

bjoto:
Cross-compilation support with riscv64 as the first target.

This PR has a dependency on https://github.com/kernel-patches/vmtest/pull/194 which also has more information.

### DIFF
--- a/build-samples/build_samples.sh
+++ b/build-samples/build_samples.sh
@@ -15,6 +15,23 @@ if [ $? -eq 0 ]; then
 	export LLVM="-$LLVM_VER"
 fi
 
+if [[ $(uname -m) != "$TARGETARCH" ]]; then
+	# Cross-compiling
+	linuxarch="$TARGETARCH"
+	case "$TARGETARCH" in
+	riscv64)
+		linuxarch="riscv"
+		;;
+	aarch64)
+		linuxarch="arm64"
+		;;
+	*)
+		;;
+	esac
+	export ARCH="$linuxarch"
+	export CROSS_COMPILE="$TARGETARCH-linux-gnu-"
+fi
+
 foldable start build_samples "Building samples with $TOOLCHAIN"
 
 if [[ "${KERNEL}" = 'LATEST' ]]; then
--- a/rootfs/Makefile
+++ b/rootfs/Makefile
@@ -8,4 +8,5 @@ all:
 			sudo ./mkrootfs_debian.sh --arch "$${arch}" --distro "$${distro}" ; \
 		done \
 	done
+	sudo ./mkrootfs_ubuntu.sh --arch riscv64 --distro jammy
 
--- a/rootfs/mkrootfs_ubuntu.sh
+++ b/rootfs/mkrootfs_ubuntu.sh
@@ -1,0 +1,180 @@
+#!/bin/bash
+# This script builds an Ubuntu root filesystem image for testing libbpf in a
+# virtual machine. Requires debootstrap >= 1.0.95 and zstd.
+
+# Use e.g. ./mkrootfs_debian.sh --arch=s390x to generate a rootfs for a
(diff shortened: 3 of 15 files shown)


rootfs/Makefile
```diff
@@ -8,4 +8,5 @@ all:
 			sudo ./mkrootfs_debian.sh --arch "$${arch}" --distro "$${distro}" ; \
 		done \
 	done
+	sudo ./mkrootfs_ubuntu.sh --arch riscv64 --distro jammy
```
chantra, line 11: why not just aligning on the distro? As ,uch as CI will be concerned, having different rootfs distro will be a nightmare. There is #84 which is a separate item to handle more rootfs distros.

rootfs/mkrootfs_ubuntu.sh
```diff
@@ -0,0 +1,180 @@
+#!/bin/bash
```
chantra, line 1: I believe `mkrootfs_debian.sh` can handle creating jammy distro just fine. I don't think this script is needed.

build-samples/build_samples.sh
```diff
@@ -15,6 +15,23 @@ if [ $? -eq 0 ]; then
 	export LLVM="-$LLVM_VER"
 fi
 
+if [[ $(uname -m) != "$TARGETARCH" ]]; then
```
chantra, line 18: I don't think TARGETARCH has been assigned yet. Probably missing TARGETARCH=$4 earlier in the script.

build-samples/build_samples.sh
```diff
@@ -15,6 +15,23 @@ if [ $? -eq 0 ]; then
...
+	aarch64)
+		linuxarch="arm64"
+		;;
+	*)
+		;;
+	esac
+	export ARCH="$linuxarch"
+	export CROSS_COMPILE="$TARGETARCH-linux-gnu-"
+fi
+
```
chantra, line 34: This is probably worth putting in helpers.sh so it can be re-used across the different actions uniformally.

chantra (commented):
Hi @bjoto , sorry for the late update on this one.

This will likely need to be tested with vmtest against your libbpf/ci repo to make sure it works, will comment there separately.

I know this is still a draft, but regardless here are some feedback.

bjoto:
> Hi @bjoto , sorry for the late update on this one.

No worries; As you noted, it's a draft, but I'm still very happy to get comments; Especially general ones!

> This will likely need to be tested with vmtest against your libbpf/ci repo to make sure it works, will comment there separately.
> 
> I know this is still a draft, but regardless here are some feedback.

ACK, and thank you.

rootfs/Makefile
bjoto: You mean align the other builds to use Ubuntu (Jammy or later)? RISC-V is still missing stable Debian support, so currently Ubuntu Jammy and later is the only option for RV. 

I agree that having different distros is not desirable.

chantra: > You mean align the other builds to use Ubuntu (Jammy or later)? RISC-V is still missing stable Debian support, so currently Ubuntu Jammy and later (or Debian Sid) is the only option for RV.
> 
> I agree that having different distros is not desirable.

That would work once we move the runners to Jammy, which is not the case.
At the moment, we build for Debian Bullseye on Ubuntu Focal.... it happens to work, but this is by luck. Jammy target would not work. The problem is due to build dependencies changes between host and target OS. #83 explains the problem.

chantra:
@bjoto I have renamed `master` branch to `main` and will delete `master`. By doing so, this PR will be closed and it does not seem there is a way to change the PR target branch. Please re-open against main at your convenience.

chantra:
relaying this here: https://github.com/kernel-patches/vmtest/pull/194#issuecomment-1947033578

PR #92, opened 2023-03-21
Add inputs.llvm-version

liu-song-6:
Since we always build BPF programs with clang, ci also installs llvm when inputs.toolchain is gcc. llvm_version(). Instead of parsing llvm version from inputs.toolchain, add a separate input field of llvm-version.

### DIFF
--- a/build-linux/build.sh
+++ b/build-linux/build.sh
@@ -10,9 +10,9 @@ ARCH="$1"
 TOOLCHAIN="$2"
 export KBUILD_OUTPUT="$3"
 
-LLVM_VER="$(llvm_version $TOOLCHAIN)" && :
-if [ $? -eq 0 ]; then
-	export LLVM="-$LLVM_VER"
+if [ $TOOLCHAIN = "llvm" ]; then
+	export LLVM="-$LLVM_VERSION"
+	TOOLCHAIN="llvm-$LLVM_VERSION"
 fi
 
 foldable start build_kernel "Building kernel with $TOOLCHAIN"
--- a/build-samples/build_samples.sh
+++ b/build-samples/build_samples.sh
@@ -10,9 +10,9 @@ KERNEL="$1"
 TOOLCHAIN="$2"
 export KBUILD_OUTPUT="$3"
 
-LLVM_VER="$(llvm_version $TOOLCHAIN)" && :
-if [ $? -eq 0 ]; then
-	export LLVM="-$LLVM_VER"
+if [ $TOOLCHAIN = "llvm" ]; then
+	export LLVM="-$LLVM_VERSION"
+	TOOLCHAIN="llvm-$LLVM_VERSION"
 fi
 
 foldable start build_samples "Building samples with $TOOLCHAIN"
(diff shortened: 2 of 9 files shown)


build-samples/build_samples.sh
```diff
@@ -10,9 +10,9 @@ KERNEL="$1"
 TOOLCHAIN="$2"
 export KBUILD_OUTPUT="$3"
 
-LLVM_VER="$(llvm_version $TOOLCHAIN)" && :
-if [ $? -eq 0 ]; then
-	export LLVM="-$LLVM_VER"
+if [ $TOOLCHAIN = "llvm" ]; then
```
chantra, line 13: same comment as above.
I believe
```
export LLVM="-${LLVM_VERSION}"
```

is enough.

We should probably add a comment explaining what LLVM is used for (see comment above) as it is non-trivial otherwise.

build-samples/build_samples.sh
```diff
@@ -25,13 +25,13 @@ fi
...
 make headers_install
 make \
-	CLANG=clang-${LLVM_VER} \
-	OPT=opt-${LLVM_VER} \
-	LLC=llc-${LLVM_VER} \
-	LLVM_DIS=llvm-dis-${LLVM_VER} \
-	LLVM_OBJCOPY=llvm-objcopy-${LLVM_VER} \
-	LLVM_READELF=llvm-readelf-${LLVM_VER} \
-	LLVM_STRIP=llvm-strip-${LLVM_VER} \
+	CLANG=clang-${LLVM_VERSION} \
```
chantra, line 28: given that `LLVM` is already `-${LLVM_VERSION}`, we may just reaplce all the references below to something similar to:
` CLANG=clang${LLVM}`

build-selftests/build_selftests.sh
```diff
@@ -10,9 +10,9 @@ KERNEL="$1"
 TOOLCHAIN="$2"
 export KBUILD_OUTPUT="$3"
 
-LLVM_VER="$(llvm_version $TOOLCHAIN)" && :
-if [ $? -eq 0 ]; then
-	export LLVM="-$LLVM_VER"
+if [[ $TOOLCHAIN = llvm* ]]; then
```
chantra, line 13: same as above.

helpers.sh
```diff
@@ -47,21 +47,6 @@ llvm_latest_version() {
   echo "17"
 }
 
-# $1 - toolchain name
```
chantra, line 50: given that you have provided the default everywhere, this may be safe to remove directly. BUT I think you would need to  change the action `get-llvm-version` to return that default value because we will not be able to land libbpf/ci changes and vmtest changes (which consume those actions) at the same time.

chantra (commented):
Thanks @liu-song-6 .

Got some comments which I think will make the change actually easier. But otherwise LGTM

build-linux/build.sh
```diff
@@ -10,9 +10,9 @@ ARCH="$1"
 TOOLCHAIN="$2"
 export KBUILD_OUTPUT="$3"
 
-LLVM_VER="$(llvm_version $TOOLCHAIN)" && :
-if [ $? -eq 0 ]; then
-	export LLVM="-$LLVM_VER"
+if [ $TOOLCHAIN = "llvm" ]; then
```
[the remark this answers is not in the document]
liu-song-6: This won't work. If we set `LLVM`, `make` command will use LLVM to build the kernel. So we cannot do this when the toolchain is gcc.

helpers.sh
liu-song-6: Can we break it for a few minutes...?

liu-song-6: Actually, `get-llvm-version` still works in current version. So I think we are good? We do change the meaning of `toolchain` though.

build-linux/build.sh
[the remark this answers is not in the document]
chantra: oh, you are right, `llvm_version` returns 1 when not using `llvm` toolchain.

build-selftests/build_selftests.sh
chantra: We have this logic in 3 different paces (and should probably move this into a function at a later stage), but this is slightly different:
`if [[ $TOOLCHAIN = llvm* ]]; then`
vs
`if [ $TOOLCHAIN = "llvm" ]; then`

is this expected? I would assume this is because the way we call the action is different. Is there any ways to uniformize it?

helpers.sh
chantra: correct. It is calling `llvm_default_version`, which is still here.

> We do change the meaning of toolchain though.

I think this is fine. We don't do a great job at naming things currently, and should provably fix this. I think it makes sense that `toolchain` becomes `llvm|gcc` and `llvm_version` the actual version.

build-selftests/build_selftests.sh
liu-song-6: oops, it should all be ` $TOOLCHAIN = "llvm" `.

PR #105, opened 2023-09-18
ansible: Install atop on Debian based hosts and default to 10s intervals

chantra:
atop is useful to be able to debug what happened historically.


chantra:
Making this a draft for now as it seems atop is not functional after installing it until I `strace` it, then it starts writtingdata to disk.

ansible/roles/base/tasks/setup-Debian.yml
```diff
@@ -16,3 +16,45 @@
...
+    state: stopped
+    enabled: false
+
+- name: Setup atop
+  become: true
+  ansible.builtin.copy:
+    dest: "/etc/default/atop"
+    content: |
+      # This file is generated by Ansible, do not modify
+      LOGOPTS="-R"
```
danielocfb, line 38: Can we use a long option?

danielocfb (approved):
Seems reasonable to me. Thanks! Any plans to export the information gathered or would we need to log into the runners to retrieve it? Would be great if we could document that somewhere, for those not familiar with `atop`.

PR #110, opened 2023-12-12
Skip log deletion on error

dubeyabhishek:
skip removing root dir in case of error to
preserve debootstrap.log

PR #111, opened 2023-12-12
rootfs: add ppc64le rootfs

chantra:
Follow-up from #110


dubeyabhishek:
@chantra I have added qemu logs using rootfs on #110

dubeyabhishek:
@chantra could you also generate a vmlinux-5.5.0 for ppc64le?

chantra:
> @chantra could you also generate a vmlinux-5.5.0 for ppc64le?

Why do you need it? We don't provide that one for s390x. Isn't testing against a local build of latest good enough? 

I am not even sure where/how those were originally created. Maybe @anakryiko has context?

PR #113, opened 2023-12-13
Debian-Gnu name lookup issue for ppc64el

dubeyabhishek:
The usr/share/dpkg/cputable contains mapping ppc64el->powerpc64le, so mkrootfs_debian.sh does lookup with qemu-powerpc64le-static, which is incorrect.


chantra:
Indeed. I cheated yesterday and created a symlink on my local FS.

chantra:
Thanks for the PR

Issue #103, opened 2023-08-17
libbpf dependency of iproute2 is too old


zhaiyan920:
Hi,

  I was encountering some relocation issue when test PR on BPF-CI. After some digging the error says:

`libbpf: prog 'test_lwt_redirect_in': bad map relo against '.rodata.str1.1' in section '.rodata.str1.1'` 

It turns out the latest root image libbpf-vmtest-rootfs-2022.10.23-bullseye-amd64.tar.zst has iproute2 with libbpf-0.3, and support for .rodata.<suffix> was introduced in commit 5bf62459b185 ("libbpf: Support multiple .rodata.* and .data.* BPF maps") from v0.6.0. Unpacking and running the root image locally with same BPF object confirms that upgrading libbpf0 would fix the issue.

Will it be feasible if admins can update the root image to pull libbpf0 from bullseye-backport? It does have slight newer version https://packages.debian.org/bullseye-backports/libbpf0, or better if we can have newer iproute2 and libbpf1 in the root image?

thanks

chantra:
Thanks @zhaiyan920  for reporting this, and sorry for getting to it so late.

In the future, we may not need to use the rootfs as part of the CI, so this should be resolved at the same time.

jerryz920:
It's all good as I worked around the issue with some variant implementation. Just curious, what alternatives do you plan to replace rootfs? Is it possible we use 9p and virtio instead in future?

chantra:
> It's all good as I worked around the issue with some variant
> implementation. Just curious, what alternatives do you plan to replace
> rootfs? Is it possible we use 9p and virtio instead in future?
>

Yeah, that would be the plan. So using the host root.

PR #117, opened 2024-01-12
run-vmtest: action to test bpf using danobi/vmtest

chantra:
The first commit message explains the rationale. Copied below.

Note that this PR only adds a new action and is essentially a no-op until kernel-patches/vmtest is modified.
## Test done:
Pointed kernel-patches/vmtest to use chantra/libbpf-ci/run-vmtest@vmtest (kernel-patches/vmtest#243): 
https://github.com/kernel-patches/vmtest/actions/runs/7506022729/job/20436527154?pr=243

To validate Meta's veristat, hacked up a PR on kernel-patches/bpf#6245:

https://github.com/kernel-patches/bpf/actions/runs/7496945955/job/20409897434?pr=6245

## First commit message:

Add a new action to mimick what used to be done by `run-qemu`.

This action is roughly a `cp run-qemu run-vmtest` in term of functionalities.
See end of this commit message for a rationale of this change.

Just like `run-qemu` assumes the presence of a rootfs which is provisioned with
a copy of the script from `ci/vmtest/run_selftests.sh`, `run-vmtest` assumes the
presence of `ci/vmtest/vmtest_selftests.sh` and will run it.

`vmtest_selftests.sh` functionally does the same as `run_selftests.sh` with a
few adjustments to make it work in the `vmtest` world` and leave in the "callee"
repository, not libbpf-ci.

`print_test_summary.py` was copied over unchanged. Later diff will remove the
version from `run-qemu` and point to this one instead.

`action.yml` needs to install a few tools that were historically baked in the rootfs.
A bunch of parameters are now historical.... this diff does not attempt to remove them
yet. This will be address later too and will probably change as libbpf/libbpf use case
is merged in.

`run.sh` gets rid of the logic to create the `qemu` one-liner as well as the
downloading of files from within the rootfs, and adjust it to using files on
disk that were left over by the test run.

Full end-to-end testing will be done through a PR in https://github.com/kernel-patches/vmtest

== Main functional difference between `run-qemu` and `run-vmtest`.

`run-qemu` runs the kernel inside a rootfs which is isolated from the host FS,
meaning that we potentially have a host in which we build the kernel/selftests
which is different than the host we run in. Causing depedency issues, see https://github.com/libbpf/ci/pull/83 and
to some extends https://github.com/libbpf/ci/issues/103 .
We could work around this with statically built binaries, but because of the different
rootfs, we end up having to maintain rootfs *and* doing a dance of copying files
in and out of the rootfs (through the prepare-rootfs action), which is slow, pulls
a fair amount of bytes through the network...

`run-vmtest` on the other hand does mount the host rootfs (read-only) and shares the
current directory (read-write) under /mnt/vmtest.
This resolves 2 things for us.
1. We don't have libraries incompatibilitie issues anymore as the building OS
and the OS running the tests will be the same.
1. We don't need to copy files in and out of the root fs. The files are already
on the FS seen by the VM, and files dumped by the tests are directly accessible
outside the VM.

On top of this, `vmtest` also handles the peculiarities of crafting the right
qemu one-liner.


chantra:
I have tested this again and it works as expected from my manual tests.

Currently using a fork of danobi/vmtest due to https://github.com/danobi/vmtest/issues/40 which is affecting s390x.

From the look of it, using vmtest instead of our old mechanism saves us 170s (p90) when testing s390x as we do not need to prepare the chroot anymore.

PR #120, opened 2024-02-13
ansible: Run qemu-static-user on all hosts

chantra:
We used to only run this role on s390x (because it needs it to be able to run x86 actions-runner on s390x).
I am planning to also leverage this for some cross-compilation work where some step unfortunately need to be able to run a binary in the target format.

Test:
```
ansible-playbook -i ~/inventory.yml ./playbook.yml   --limit aws -t qemu
```


chantra:
I pushes  this to hosts and no issue came up. Landing.

PR #121, opened 2024-02-13
actions: Upgrade checkout action to v4

chantra:
v3 depends on nodejs16 which is being deprecated. Upgrade to v4.


danielocfb (approved):
Looks good to me.

chantra:
Ran some workflow with this and no problem to be seen.... landing.

PR #122, opened 2024-02-13
cross-compilation: make libbpf/ci cross-compilation aware

chantra:
This adds handling of a required `arch` parameter.

setup-build-env, build-linux, build-sample, build-selftests are made to handle this input and set the `ARCH` variable when calling `make` to what is expected by Linux build system.

If the `inputs.arch` is different than the host arch, `CROSS_COMPILE` variable will be set to the right toolchain.

`setup-build-env` is also modified to install the corresponding target libraries.

### DIFF
--- a/build-linux/build.sh
+++ b/build-linux/build.sh
@@ -6,9 +6,17 @@ THISDIR="$(cd $(dirname $0) && pwd)"
 
 source "${THISDIR}"/../helpers.sh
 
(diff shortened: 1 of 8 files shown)


chantra:
Example end-2-end run https://github.com/kernel-patches/vmtest/actions/runs/7891345677/job/21537206452?pr=262

PR #123, opened 2024-04-04
ansible: don't disable auditd if the package is not installed

iii-i:
Fresh Ubuntu installs don't have auditd.


chantra:
Thanks @iii-i

PR #126, opened 2024-05-23
run-vmtest: Configure root-less usage of KVM

danobi:
For GH managed runners, KVM is generally available, but we need to configure it so that non-root users can use it. This shouldn't affect the self-hosted runners where we are already root.


danobi:
Tested in:

prod: https://github.com/kernel-patches/bpf/pull/7090
rc: https://github.com/kernel-patches/bpf-rc/pull/3885

rc has some selftest failures, but at least CI is working

PR #130, opened 2024-08-14
ci: Install libpcap for upcoming traffic monitor feature.

ThinkerYzu1:
Traffic monitor is going to be landed on upstream, capturing network traffic in the background for selected test cases. libpcap is required to enable this feature.

Also upload the archives of traffic log files so that developers can download them if necessary.

https://lore.kernel.org/all/20240810023534.2458227-1-thinker.li@gmail.com/

Should be landed after 
https://github.com/kernel-patches/vmtest/pull/280


theihor:
Closing in favor of #168

PR #134, opened 2024-08-21
Switch containers to use Ubuntu noble

chantra:
This contains a couple of steps:
1. Add the possibility to override on a per runner basis the value of DOCKER_TAG by setting its value in /etc/action-runner/runner_unit-\d\d.env
2. make the containers advertise a new docker-{distro}-main instead of docker-main . Current actions are looking for tag docker-main, so by using this new value, the containers won't be picked up
3. change the value of DOCKER_TAG in /etc/action-runner/runner_unit.env This will make ALL container pick the ubuntu noble container version, and should be deployed only once all actions are running on ubuntu noble.


chantra:
This has been pushed to prod a while now. I am landing this to be able to fully retire old setup.

PR #143, opened 2024-09-12
Factor out incremental build prep into an action

### LINKED ISSUE #142: packages vmtest steps into libbpf/ci actions
We have some steps/actions in kernel-patches/vmtest that are not fully encapsulated:

```
.github/actions/veristat_baseline_compare/action.yml:      run: ./.github/scripts/compare-veristat-results.sh
.github/scripts/compare-veristat-results.sh:python3 ./.github/scripts/veristat_compare.py compare.csv
.github/scripts/tar-artifact.sh:    "scripts/"
.github/workflows/kernel-build.yml:          bash .github/scripts/get-commit-metadata.sh
.github/workflows/kernel-build.yml:          bash .github/scripts/prepare-incremental-builds.sh ${{ steps.get-commit-metadata.outputs.commit }}
.github/workflows/kernel-build.yml:          bash .github/scripts/tar-artifact.sh ${{ inputs.arch }} ${{ inputs.toolchain_full }}
```
Because the actions/workflow use local copies of the scripts, they can only be used by the kernel-patches/vmtest (or anything that copies the files locally). Running this from another repo would fail to find those scripts.

We should package those individual steps into libbpf/ci actions which we then can re-use from kernel-patches/vmtest and/or libbpf/libbpf.

theihor:
Partially resolves #142 

@chantra I've been testing this change in a [simplified workflow](https://github.com/theihor/libbpf-ci/actions/workflows/simplified-kbuild.yml) on my fork.

Any suggestions how to verify [kernel-patches/bpf ](https://github.com/kernel-patches/bpf) CI would work with the change? I guess we could just push and see, but maybe there is a less disruptive way.


chantra:
Thanks @theihor 

I have not fully reviewed the whole PR yet, but from  quick pass, it seems fine so far.

> Any suggestions how to verify [kernel-patches/bpf ](https://github.com/kernel-patches/bpf) CI would work with the change? I guess we could just push and see, but maybe there is a less disruptive way.

Yes, you should create a PR against kernel-patches/vmtest, apply the changes would expect to do when migrating to using the new action, but until this is landed, you will need to use an action path pointing to your repo/branch, e.g 

`theihor/libbpf-ci/prepare-incremental-build@prepare-incremental-build-action`
instead of `libbpf-ci/{action}@main`.

prepare-incremental-build/action.yml
```diff
@@ -0,0 +1,52 @@
+name: 'Prepare incremental kernel build'
+description: 'Pull cached kernel build output from previous runs and prepare the repo for incremental build'
+inputs:
+  repo-root:
+    description: "Root of the kernel repository"
+    required: true
+    default: ${{ github.workspace }}
```
chantra, line 7: in my experience, this does not work. We have it in a few places, but I don't think it works.

It may be worth testing separately through a simple action.

Based on https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#context-availability , it is not mentioned that any contexts are available in the "inputs" part of an action. (It would work for a reusable workflow though).

prepare-incremental-build/action.yml
```diff
@@ -0,0 +1,52 @@
...
+  steps:
+
+    - name: Get commit meta-data for cache lookup
+      id: get-commit-metadata
+      shell: bash
+      working-directory: ${{ inputs.repo-root }}
+      run: ${GITHUB_ACTION_PATH}/get-commit-metadata.sh ${{ inputs.base-branch }}
+
+    - name: Pull recent KBUILD_OUTPUT contents
+      uses: actions/cache@v4
```
chantra, line 37: I am pretty sure we should set the working-directory here too.
I believe we are lucky it works because the caller sets `repo-root` to `github.workspace`, which happens to be the default working directory.

prepare-incremental-build/action.yml
chantra: replying to myself... on the other hand, there is things like https://github.com/actions/checkout/blob/6d193bf28034eafb982f37bd894289fe649468fc/action.yml#L6C5-L6C38 which suggests it should work, so mayke I did something wrong when testing.

chantra (commented):
I think there is potentially 1 flaw that we were lucky to not hit so far. e.g if we were to run the action outside of the github repo.

For the case of kernel-patches/bpf... the workspace just happens to be a linux repo. For other projects, we do download the sources in the step above and move it in place, so we end up being lucky and having the workspace being a linux repo too.

I believe, if we had the action working properly, we should be able to skip the "move linux source in place" step. I think this is out of scope for this task, but somethign to keep in mind for later to make the action self-contained.

Finally, lets take the opportunity when working on actions to write a README.md in the action directory to document how to use it.  Similarly to how actions in the marketplace do: https://github.com/actions/checkout/blob/main/README.md for instance.

prepare-incremental-build/action.yml
theihor: My understading is that this particular step invoking `actinons/cache@v4` does not have to be executed from the repository directory. The path to store the result of cache lookup (if it's a hit) is passed via `path` parameter:
```yaml
    - name: Pull recent KBUILD_OUTPUT contents
      uses: actions/cache@v4
      with:
        path: ${{ inputs.kbuild-output }} # <-- THIS
        key: ...
```

And we path is through as an action parameter, although my impression is it never changes for our workflows.

And then the next step (that runs `prepare-incremental-builds.sh`) expects to be executed in the repository root, but also expects KBUILD_OUTPUT env variable to point to the cached build.

```yaml
    - name: Prepare incremental build
      shell: bash
      env:
        KBUILD_OUTPUT: ${{ inputs.kbuild-output }}
      working-directory: ${{ inputs.repo-root }} # and so here we do have to set path to the repository
```

I guess it is confusing that we do not set working-directory in cache step, but do set it for other steps, but the action should work either way.

prepare-incremental-build/action.yml
theihor: Yes, it's not clear from the docs. I can double check, although in this case I think we can just remove the default and require this input to be specified.

theihor:
@chantra 

> For the case of kernel-patches/bpf... the workspace just happens to be a linux repo. For other projects, we do download the sources in the step above and move it in place, so we end up being lucky and having the workspace being a linux repo too.
> 
> I believe, if we had the action working properly, we should be able to skip the "move linux source in place" step. I think this is out of scope for this task, but somethign to keep in mind for later to make the action self-contained.

In the simplified workflow where is was testing this action, I deliberately checked out `kernel-patches/bpf` into a non-workspace directory, and then passed a path to that directory as action input (see [here](https://github.com/theihor/libbpf-ci/blob/main/.github/workflows/simplified-kbuild.yml#L60-L73)).

So, assuming I haven't missed anything, the action doesn't care how the linux source tree directory was created: via checkout or via "download linux source" steps. As long as the correct path to the git repo is passed, the action will try finding the cached build output.

> Finally, lets take the opportunity when working on actions to write a README.md in the action directory to document how to use it. Similarly to how actions in the marketplace do: https://github.com/actions/checkout/blob/main/README.md for instance.

Will do.

prepare-incremental-build/action.yml
chantra: I am not convinced `path` work as you are explaining it. Based on https://github.com/actions/cache?tab=readme-ov-file#inputs

my understanding is that it will extract the "kbuild-output" directory from the cache archive.

side note... we are not caching as part of libbpf/ci (e.g https://github.com/libbpf/ci/actions/caches shows no cache, we may want to also exercise this in the future.).

prepare-incremental-build/get-commit-metadata.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# branch="${GITHUB_BASE_REF}"
+
+# if [ "${GITHUB_EVENT_NAME}" = 'push' ]; then
+#   branch="${GITHUB_REF_NAME}"
+# fi
+
```
chantra, line 8: maybe remove this now that this is handled in one of the action'step.

chantra:
> In the simplified workflow where I was testing this action, I deliberately checked out `kernel-patches/bpf` into a non-workspace directory, and then passed a path to that directory as action input (see [here](https://github.com/theihor/libbpf-ci/blob/main/.github/workflows/simplified-kbuild.yml#L60-L73)).
> 
> So, assuming I haven't missed anything, the action doesn't care how the linux source tree directory was created: via checkout or via "download linux source" steps. As long as the correct path to the git repo is passed, the action will try finding the cached build output.

yeah, I think we are missing something but it silently fails.
In your case, no cache was available: https://github.com/theihor/libbpf-ci/actions/runs/10836721377/job/30071091347#step:4:63
What I am afraid is that at best we would download the cache but in an directory which is not accessible when we do the build.
It is probably easier to test/validate with a actions/cache basic workflow.

prepare-incremental-build/get-commit-metadata.sh
theihor: done in my fork

theihor:
> yeah, I think we are missing something but it silently fails.
> In your case, no cache was available: https://github.com/theihor/libbpf-ci/actions/runs/10836721377/job/30071091347#step:4:63
> What I am afraid is that at best we would download the cache but in an directory which is not accessible when we do the build.
> It is probably easier to test/validate with a actions/cache basic workflow.

In that workflow there was nothing to cache actually... I wrote another one, and there you can see the cache was
- created by https://github.com/theihor/libbpf-ci/actions/runs/10857913112/job/30135143759
- and then restored in subsequent run: https://github.com/theihor/libbpf-ci/actions/runs/10857937779/job/30135210069

Btw, do we want workflows like that (created specifically to test a custom action) in libbpf/ci? I am not sure.

prepare-incremental-build/action.yml
theihor: > I am not convinced path work as you are explaining it. Based on https://github.com/actions/cache?tab=readme-ov-file#inputs
> 
> my understanding is that it will extract the "kbuild-output" directory from the cache archive.

Maybe I explained it badly, but I think we both understand it correctly.

The cache action is designed to store things like gradle caches ([docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#input-parameters-for-the-cache-action)), so I am pretty sure its `path` input is used both for uploading to the "cloud" after job has finished, and then restoring to the same `path` later, if lookup key matches.

Although actually, if the path is relative, then working directory might be important. So to be on the safe side, I'll set working-directory to the kernel repo root.

chantra: > The cache action is designed to store things like gradle caches ([docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#input-parameters-for-the-cache-action)), so I am pretty sure its `path` input is used both for uploading to the "cloud" after job has finished, and then restoring to the same `path` later, if lookup key matches.

agreed

> Although actually, if the path is relative, then working directory might be important. So to be on the safe side, I'll set working-directory to the kernel repo root.

I think this would make the composite action work as expected *if* the kernel-root does not happen to be the github.workspace.

On the other hand, kbuild-output could be anywhere on the filesystem and we should be able to build by providing the absolute path. If we remove the need for kbuild-output to be in kernel-root, we could leverage the cache for workflows outside of kernel-patches/bpf .But that should probably go in a separate PR.

chantra:
> Btw, do we want workflows like that (created specifically to test a custom action) in libbpf/ci? I am not sure.

Maybe not using the key used by the kernel. But given the hard time we have to validate it works as expected, I think it could be useful to have some integration workflows to avoid regression in the future.

prepare-incremental-build/action.yml
theihor: It turns out that it is not possible to specify `working-directory` in a step that uses an action (as opposed to running a script). If you try, [job fails on execution](https://github.com/theihor/libbpf-ci/actions/runs/10893669399). So, in our case, passing an absolute `path` to `actions/cache` is the only good option really.

prepare-incremental-build/action.yml
theihor: Done.

theihor:
@chantra I noticed that artifacts produced by `tar-artifact.sh` containe two copies of selftests/bpf: `tools/testing/selftests/bpf` and `selftests/bpf`.

If found that one is copied into another here: https://github.com/libbpf/ci/blob/main/build-selftests/build_selftests.sh#L63-L66

Do you know if there is a reasion for having two copies?

chantra:
> @chantra I noticed that artifacts produced by `tar-artifact.sh` contain two copies of selftests/bpf: `tools/testing/selftests/bpf` and `selftests/bpf`.
> 
> I found that one is copied into another here: https://github.com/libbpf/ci/blob/main/build-selftests/build_selftests.sh#L63-L66
> 
> Do you know if there is a reason for having two copies?

No, I don't think so. The artifact is downloadable from GH (when going to the summary of the run, there is a link to the artifacts). The content of vmlinux-x86_64-gcc is https://gist.github.com/chantra/0acb0ec04a58db3adea2debd47e73fee

I think it could be moved in place.
I know that we keep the "Makefile"s so later we can run `make -s image_name` to discover where the vmlinux file is. The rest, I am not too sure. Even in the selftests/bpf directory we may not need to keep everything that is compiled. At least not for the "artifact" that we re-use during the test run. We do need to keep it for when we update the cache though.

chantra (approved):
LGTM. Before this lands, do you mind testing on the kernel-patches/vmtest repo by using this action, just to make sure we did not missed anything.

It looks like the tar logic as a whole needs to be re-assessed, but this is a separate story/change that can be followed-up later.

theihor:
> > @chantra I noticed that artifacts produced by `tar-artifact.sh` contain two copies of selftests/bpf: `tools/testing/selftests/bpf` and `selftests/bpf`.
> > I found that one is copied into another here: https://github.com/libbpf/ci/blob/main/build-selftests/build_selftests.sh#L63-L66
> > Do you know if there is a reason for having two copies?
> 
> No, I don't think so. The artifact is downloadable from GH (when going to the summary of the run, there is a link to the artifacts). The content of vmlinux-x86_64-gcc is https://gist.github.com/chantra/0acb0ec04a58db3adea2debd47e73fee
> 
> I think it could be moved in place. I know that we keep the "Makefile"s so later we can run `make -s image_name` to discover where the vmlinux file is. The rest, I am not too sure. Even in the selftests/bpf directory we may not need to keep everything that is compiled. At least not for the "artifact" that we re-use during the test run. We do need to keep it for when we update the cache though.

I believe on `kernel-patches/bpf` CI only one of those two `selftests/bpf` copies is present. For example [build job artifacts](https://github.com/kernel-patches/bpf/actions/runs/10928982702/artifacts/1950803125) from [this bpf-ci run](https://github.com/kernel-patches/bpf/actions/runs/10928982702/job/30349420670) contain only this inside: 
```bash
$ tree -L 2
.
├── kbuild-output
│   ├── arch
│   ├── include
│   └── vmlinux
└── selftests
    └── bpf
```
And the flag controlling what's included in the tarball depends on the repository name.

So the question is why additional files are needed on `kernel-patches/vmtest`, but not on `kernel-patches/bpf`. I'll try to figure that out. If it's only for `make image_name` it seems silly, as we should be able to pass that name around as an input or env variable.

> LGTM. Before this lands, do you mind testing on the kernel-patches/vmtest repo by using this action, just to make sure we did not missed anything.

I already did: https://github.com/kernel-patches/vmtest/pull/287

I tagged you there, but wasn't able to add you as a reviewer.

.github/scripts/tar-artifact.sh
```diff
@@ -39,31 +48,56 @@ find selftests/ -name "*.o" -a ! -name "*.bpf.o" -print0 | \
...
 if [ $archive_make_helpers -ne 0 ]; then
   # Package up a bunch of additional infrastructure to support running
   # 'make kernelrelease' and bpf tool checks later on.
   mapfile -t additional_file_list < <(find . -iname Makefile)
   additional_file_list+=(
-    "scripts/" 
-    "tools/testing/selftests/bpf/" 
-    "tools/include/" 
+    "scripts/"
+    "tools/testing/selftests/bpf/"
```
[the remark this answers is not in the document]
theihor: Yeah, I thought about it, but didn't want to risk breaking something, before I'm sure this is unnecessary.

I agree with your suggestion to re-assess build artifacts in a separate change, as this one is quite big already.

.github/scripts/tar-artifact.sh
```diff
@@ -39,31 +48,56 @@ find selftests/ -name "*.o" -a ! -name "*.bpf.o" -print0 | \
...
-  --exclude '*.d' \
-  --exclude '*.h' \
-  --exclude '*.output' \
-  selftests/bpf/ | zstd -T0 -19 -o "vmlinux-${arch}-${toolchain}.tar.zst"
+    kbuild-output \
+    "${additional_file_list[@]}" \
+    --exclude '*.cmd'                  \
+    --exclude '*.d'                    \
+    --exclude '*.h'                    \
+    --exclude '*.output'               \
```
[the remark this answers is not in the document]
theihor: I'll have to double check this, but IIRC `--exclude` applies to the files after it is specified.

So `tar -cf - foo --exclude="*.x" bar` will filter out `*.x` only from bar.

In our case it means that the filters don't apply to `additional_file_list`. Whether this was intentional, I don't know. Probably not.

chantra:
> So the question is why additional files are needed on kernel-patches/vmtest, but not on kernel-patches/bpf. I'll try to figure that out. If it's only for make image_name it seems silly, as we should be able to pass that name around as an input or env variable.

`make image_name` only needs the Makefiles, and I don't think it needs anything related to selftests. So I think this is probably unrelated and just ended up there as an accident. As much as this PR is concerned, I think we should leave it like this, and follow-up in a separate PR.

> I already did: [kernel-patches/vmtest#287](https://github.com/kernel-patches/vmtest/pull/287)
> 
> I tagged you there, but wasn't able to add you as a reviewer.

Ha great. Thanks. If this works, then that's a high signal that everything should be fine :D

chantra:
Feel free to merge this whenever you want, we may want to keep an eye on both kernel-patches/bpf and libbpf/libbpf after it landed just to be safe.

theihor:
> Feel free to merge this whenever you want, we may want to keep an eye on both kernel-patches/bpf and libbpf/libbpf after it landed just to be safe.

Ok. Thank you for the review.

I'd like BPF CI to become stable before merging this.

PR #144, opened 2024-09-16
add BPF GCC support for BPF CI

cupermir:
This PR adds support for building BPF GCC for the purpose to integrate it in BPF CI.

Similar support that makes use of this code will be pull requested in kernel-patches/bpf repo.

### DIFF
--- a/build-selftests/build_selftests.sh
+++ b/build-selftests/build_selftests.sh
@@ -10,9 +10,11 @@ TARGET_ARCH="$1"
 KERNEL="$2"
 TOOLCHAIN="$3"
 export KBUILD_OUTPUT="$4"
+BPF_TOOLCHAIN="$5"
 
 ARCH="$(platform_to_kernel_arch ${TARGET_ARCH})"
 CROSS_COMPILE=""
+BPF_GCC_BIN="${GITHUB_WORKSPACE}/install_bpf_gcc/bin/bpf-unknown-none-gcc"
 
 if [[ "${TARGET_ARCH}" != "$(uname -m)" ]]
 then
@@ -51,6 +53,12 @@ MAKE_OPTS=$(cat <<EOF
 	VMLINUX_H=${VMLINUX_H}
 EOF
 )
+
+if [[ "${BPF_TOOLCHAIN}" == 'gcc' ]]
+then
+	MAKE_OPTS="${MAKE_OPTS} BPF_GCC=${BPF_GCC_BIN}"
+fi
+
 SELF_OPTS=$(cat <<EOF
 	-C ${REPO_ROOT}/${REPO_PATH}/tools/testing/selftests/bpf
 EOF
--- a/setup-build-env/install_bpf_gcc.sh
+++ b/setup-build-env/install_bpf_gcc.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+set -eu
+
+source $(cd $(dirname $0) && pwd)/../helpers.sh
+
+LOGFILE=build-gnu-bpf.log
+INSTALLDIR=${GITHUB_WORKSPACE}/install_bpf_gcc/
+
+mkdir -p build_bpf_gcc
+cd build_bpf_gcc
+
+foldable start install_bpf_gcc "Installing BPF GCC"
+BINUTILS_TARBALL=`wget https://snapshots.sourceware.org/binutils/trunk/latest/src/sha512.sum -O - -o /dev/null | grep -E 'binutils-[0-9a-f.-]+.tar.xz' | sed -e 's/.*\(binutils-[^<]*\).*/\1/'`
+GCC_TARBALL=`wget https://gcc.gnu.org/pub/gcc/snapshots/LATEST-15 -O - -o /dev/null | grep -E 'gcc-15-[0-9]+.tar.xz' | sed -e 's/.*\(gcc-15-[^<]*\).*/\1/'`
+
+BINUTILS_URL="https://snapshots.sourceware.org/binutils/trunk/latest/src/$BINUTILS_TARBALL"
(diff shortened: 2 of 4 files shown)


setup-build-env/install_bpf_gcc.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+set -eu
+
+source $(cd $(dirname $0) && pwd)/../helpers.sh
+
+LOGFILE=build-gnu-bpf.log
+INSTALLDIR=$(pwd)/install_bpf_gcc/
```
chantra, line 7: I have mixed filling about using `pwd` here. It is probably better to stick to `${GITHUB_WORKSPACE}` here so it is consistently at the same location and not dependent on the current working directory the workflow is running the action in.

build-selftests/build_selftests.sh
```diff
@@ -10,9 +10,11 @@ TARGET_ARCH="$1"
 KERNEL="$2"
 TOOLCHAIN="$3"
 export KBUILD_OUTPUT="$4"
+BPF_TOOLCHAIN="$5"
 
 ARCH="$(platform_to_kernel_arch ${TARGET_ARCH})"
 CROSS_COMPILE=""
+BPF_GCC_BIN="$(pwd)/install_bpf_gcc/bin/bpf-unknown-none-gcc"
```
chantra, line 17: same here regarding `$(pwd)`

chantra (changes requested):
Thanks @cupermir .

Overall, LGTM. In order to be able to test this self-isolate, would you mind adding an entry in https://github.com/libbpf/ci/blob/f1bc702dc76510c4b306a76d1a00db5dc29fb99b/.github/workflows/test.yml#L20 to exercise the gcc bpf-toolchain install/run?

I am actually wondering if this will uncover that it is also needed as part of the testing environment? In which case, during the test run we would need to install the toolchain again.

setup-build-env/install_bpf_gcc.sh
cupermir: Sure.
Please notice that the `source $(cd $(dirname $0) && pwd)/../helpers.sh` is already in setup-build-env/install_clang.sh.

chantra: Yes. This is fine. I meant this comment for INSTALL_DIR. The former is basically to build a path to the helper file in repo, regardless of the working directory.

cupermir:
> https://github.com/libbpf/ci/blob/f1bc702dc76510c4b306a76d1a00db5dc29fb99b/.github/workflows/test.yml#L20
> to exercise the gcc bpf-toolchain install/run?
> 
> I am actually wondering if this will uncover that it is also needed as part of the testing environment? In which case, during the test run we would need to install the toolchain again.

If I understand well what is happening in test.yml, there is nothing to be added to it, since the bpf-gcc this PR instroduces is only a bpf compiler, it would not impact kernel building in any way.

setup-build-env/install_bpf_gcc.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+set -eu
+
+source $(cd $(dirname $0) && ${GITHUB_WORKSPACE})/../helpers.sh
```
chantra, line 4: this was correctly using `$(pwd)`. This line essentially used to be able to find the `helpers.sh` file in the parent directory of this script. There is no warranty, and likely is is never going to be, in the parent directory of `GITHUB_WORKSPACE`.

Currently, this code is not ran because we do not `bpf-toolchain` is not set in the test cases.

chantra (commented):
> > https://github.com/libbpf/ci/blob/f1bc702dc76510c4b306a76d1a00db5dc29fb99b/.github/workflows/test.yml#L20
> > 
> > to exercise the gcc bpf-toolchain install/run?
> > I am actually wondering if this will uncover that it is also needed as part of the testing environment? In which case, during the test run we would need to install the toolchain again.
> 
> If I understand well what is happening in test.yml, there is nothing to be added to it, since the bpf-gcc this PR instroduces is only a bpf compiler, it would not impact kernel building in any way.

The test.yml file here builds (kernel and selftests) and run the tests. If your diff was adding an the right bpf-toolchain metadata to the toolchain matrix, you could get the testing to actually exercise building selftests with GCC.

the test.yml file is calling "reusable" workflows that thenselves take inputs, which are provided in https://github.com/libbpf/ci/blob/main/.github/workflows/test.yml#L33 and used in the reusable workflow https://github.com/libbpf/ci/blob/main/.github/workflows/kernel-build-test.yml (which IIRC, also calls other reusable workflow).

setup-build-env/install_bpf_gcc.sh
cupermir: Just patched the pwd regression.

> Currently, this code is not ran because we do not `bpf-toolchain` is not set in the test cases.

Honestly, I would not know where to start to add it in.

chantra: https://github.com/libbpf/ci/blob/f1bc702dc76510c4b306a76d1a00db5dc29fb99b/.github/workflows/test.yml#L20-L23

This is where you would want to modify the current entries to have the bpf-toolchain set to whatever the default is, and add an entry where bpf-toolchain is set to GCC.

You will then need to pass the toolchain through the with statement: https://github.com/libbpf/ci/blob/f1bc702dc76510c4b306a76d1a00db5dc29fb99b/.github/workflows/test.yml#L33

and probably need to propagate it in the reusable workflows that are indirectly used.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#using-a-matrix-strategy may help in understanding how this works.

chantra: IIRC, you also need to run a separate set of tests for gcc, so you may need more tweaking.

theihor:
Closing in favor of #164

PR #145, opened 2024-10-11
ansible: mask systemd-binfmt service on qemu-user-static install

theihor:


ansible/roles/qemu-user-static/tasks/main.yml
```diff
@@ -23,6 +23,14 @@
     group: root
   register: qemu_user_static
 
+- name: Mask systemd-binfmt service
+  become: true
+  systemd:
+    name: systemd-binfmt
+    state: stopped
+    masked: yes
+  register: mask_systemd_binfmt
```
chantra, line 32: I don't think this is needed if you don't use it later: https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#registering-variables

chantra (approved):
not a big deal, but you can probably remove that register keyword.

chantra:
Also adding more details to the patch/PR description will help a potential future person to understand why the decision was made to mask that service.

ansible/roles/qemu-user-static/tasks/main.yml
theihor: Removed.

theihor:
> Also adding more details to the patch/PR description will help a potential future person to understand why the decision was made to mask that service.

Updated commit message. Please let me know if I missed anything important.

chantra:
LGTM. Feel free to land whenever.

PR #146, opened 2024-10-16
Add build-scx-selftests action

theihor:
Add `build-scx-selftests` action. This action will build `tools/testsing/selftests/sched_ext` binaries, given kbuild-output and toolchain parameters passed to the action.

The action and its build script are very similar to `build-selftests`. Build scipt is simplified somewhat.

Action inputs are different:
* `repo-root` is a new required parameter, because we don't want to rely on the assuption that `$REPO_ROOT` == `github.workspace`
* `kernel` is removed: it controled the location of `vmlinux.h`, but it is generated from the `$KBUILD_OUTPUT/vmlinux`, and I don't think we need to be able to customize it (at least not yet)
    * https://github.com/kernel-patches/bpf/blob/bpf-next_base/tools/testing/selftests/sched_ext/Makefile#L125-L131
    * https://github.com/theihor/libbpf-ci/blob/build-scx/build-scx-selftests/build.sh#L30

### DIFF
--- a/.github/scripts/tar-artifact.sh
+++ b/.github/scripts/tar-artifact.sh
@@ -13,7 +13,6 @@ fi
 
 arch="${1}"
 toolchain="${2}"
-archive_make_helpers="${3:-0}"
 
 # Convert a platform (as returned by uname -m) to the kernel
 # arch (as expected by ARCH= env).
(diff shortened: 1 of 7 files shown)


theihor:
WIP vmtest change: https://github.com/kernel-patches/vmtest/pull/294

chantra:
LGTM. Can you add it as a step in https://github.com/libbpf/ci/blob/f605c6229129f5abeade6c40fc7bcd4bb6e19587/.github/workflows/kernel-build.yml#L108-L120 so we can make sure we are building/packing the content successfully.

PR #147, opened 2024-10-17
Reorganize reusable workflows

theihor:
We want to maintain only one copy of each reusable workflow between libbpf/ci and kernel-patches/vmtests [1]. To achieve that, libbpf/ci is going to be the source of truth for them, and kernel-patches/vmtest main workflow will be calling libbpf/ci.

This change reorganizes files to make it clear which workflows are reusable, and which are repository-specific. I suggest the following convention:

* `.github` contains workflows local to the owner repository
* `workflows` contains reusable workflows, intened for external use

[1]: https://github.com/kernel-patches/vmtest/tree/master/.github


theihor:
As it turns out, we can't do that [because reasons](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#creating-a-reusable-workflow). Not only can't we move to repo root, we also can't use a subdirectory in `.github/workflows`.

> Reusable workflows are YAML-formatted files, very similar to any other workflow file. As with other workflow files, you locate reusable workflows in the **.github/workflows** directory of a repository. **Subdirectories of the workflows directory are not supported.**

PR #148, opened 2024-10-21
Factor out tar-artifacts action

theihor:
Rework tar-artifact.sh script, making it suitable for use as an independent github action.

Use tar -rf (append) to accumulate files of interest before compressing with zstd.

Control groups of artifacts to include (e.g. selftests/bpf, selftests/sched_ext) with environment variables.


.github/workflows/kernel-build.yml
```diff
@@ -140,9 +139,17 @@ jobs:
...
           llvm-version: ${{ inputs.llvm-version }}
       - name: Tar artifacts
-        working-directory: ${{ env.REPO_ROOT }}
-        run: |
-          bash .github/scripts/tar-artifact.sh ${{ inputs.arch }} ${{ inputs.toolchain_full }}
+        id: tar-artifacts
+        uses: ./tar-artifacts
+        env:
+          ARCHIVE_BPF_SELFTESTS: 'true'
+          ARCHIVE_MAKE_HELPERS: ${{ github.repository != 'kernel-patches/bpf' && 'true' || '' }}
```
chantra, line 146: thought for later: maybe we should move this out of libbpf/ci eventually and let the caller set it through an env var or input (probably preferred)

tar-artifacts/action.yml
```diff
@@ -0,0 +1,28 @@
...
+    description: "Path to the root of the kernel repository"
+    required: true
+  arch:
+    required: true
+  toolchain:
+    required: true
+outputs:
+  archive-name:
+    description: 'Artifacts tarball name'
+    value: ${{ steps.run-script.outputs.archive-name }}
```
chantra, line 16: toolchain is only needed so we can generate the final archive name. I wonder if we should rather provide the destination file as an input, this way we can give it the name we like from the caller.

It also seems to be what other actions doing something similar would do: https://github.com/marketplace/actions/tar-action

chantra (approved):
tar -r trick is nice to separate each step. I suppose this could increase IO compare to tar | zstd, but at the same time, we used to do some copies left and right before, so probably on par with what we had.

.github/workflows/kernel-build.yml
theihor: Right, this is why I made them env vars in the first place. They are set in the workflow step right now, I guess you're talking about a case when the workflow itself is called from outside.

When we figure out how to actually use reusable workflows in kernel-patches/vmtest, this might be worth revisiting.

tar-artifacts/action.yml
theihor: Added `archive` input and removed output.

I thought about this, and wasn't sure what's the best API would be. The caller obviously needs to know the archive name, so why don't just allow them to set it.

Also, it seems so far that using env variables to control action-driven scripts is very convenient. We should think about a convention for when to use env variables vs script argument vs action inputs, because right now it's quite chaotic.

chantra: > because right now it's quite chaotic.

could not agree more :D

chantra (approved):

PR #149, opened 2024-10-24
run-vmtest: sched_ext selftests support

theihor:


theihor:
@chantra I think we can merge this in

When we are ready to enable scx test runs, we'll just need to uncomment [this line](https://github.com/libbpf/ci/pull/149/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R30).

run-vmtest/run.sh
```diff
@@ -46,12 +46,11 @@ foldable start vmtest "Starting virtual machine..."
 
 # Tests may be comma-separated. vmtest_selftest expect them to come from CLI space-separated.
 T=$(echo ${KERNEL_TEST} | tr -s ',' ' ')
-# HACK: We need to unmount /tmp to access /tmp from the container....
-vmtest -k "${VMLINUZ}" --kargs "panic=-1 sysctl.vm.panic_on_oom=1" "umount /tmp && \
```
chantra, line 36: I was surprised to see this.... but we essentially do not need to umount /tmp since https://github.com/danobi/vmtest/commit/0a63dce0e073273ba79c8dd4bfba10b571bcbd09

.github/workflows/test.yml
```diff
@@ -20,7 +20,13 @@ jobs:
         toolchain:
           - {"name": "gcc", "fullname": "gcc", "version": 17}
           - {"name": "llvm", "fullname": "llvm-17", "version": 17}
-        tests: [{"include": [{"test": "test_progs", "continue_on_error": false, "timeout_minutes": 360}, {"test": "test_progs_no_alu32", "continue_on_error": false, "timeout_minutes": 360}, {"test": "test_verifier", "continue_on_error": false, "timeout_minutes": 360}, {"test": "test_maps", "continue_on_error": false, "timeout_minutes": 360}]}]
+        tests:
```
chantra, line 23: nice!

run-vmtest/run.sh
```diff
@@ -3,14 +3,28 @@
 set -euo pipefail
 trap 'exit 2' ERR
 
-source $(cd $(dirname $0) && pwd)/../helpers.sh
+source "${GITHUB_ACTION_PATH}/../helpers.sh"
```
chantra, line 6: why the change from `$(cd $(dirname $0) && pwd)` to `${GITHUB_ACTION_PATH}`?

run-vmtest/run.sh
```diff
@@ -3,14 +3,28 @@
...
+
+if [[ "$KERNEL_TEST" != "sched_ext" ]]; then
+	VMTEST_SCRIPT="${GITHUB_ACTION_PATH}/../ci/vmtest/vmtest_selftests.sh"
+	if [[ "${KERNEL}" = 'LATEST' ]]; then
+		RUN_BPFTOOL_CHECKS=true
+	fi
+else
+	VMTEST_SCRIPT="${GITHUB_ACTION_PATH}/../ci/vmtest/sched_ext_selftests.sh"
+fi
+
```
chantra, line 18: how about we use `KERNEL_TEST` has the key to find which vmtest script to use?
Probably worth renaming `vmtest_selftests.sh` into `bpf_selftests.sh`? And use this as a key?

chantra (commented):
LGTM. I wonder if another diff on top would be worthwhile to uniformize how we discover which `VMTEST_SCRIPT` to use.

chantra (approved):

run-vmtest/run.sh
chantra: it seems you dealt with this in #150 :)

run-vmtest/run.sh
theihor: Yeah. I actually **had to** remove it, otherwise it failed to umount.

run-vmtest/run.sh
theihor: I've been trying to use common env vars for paths everywhere, and `$GITHUB_ACTION_PATH` makes most sense for actions, when you're looking for relative paths.

However, in this particular case I don't think important, so I don't mind leaving pwd.

Any particular reasons in favor of `$(cd $(dirname $0) && pwd)`?

theihor:
> LGTM. I wonder if another diff on top would be worthwhile to uniformize how we discover which `VMTEST_SCRIPT` to use.

Thanks.

There are more changes to it in #150, and follow ups are very likely. Initially I thought to pass VMTEST_SCRIPT as an action input, and then changed my mind (don't remember why). Maybe we'll do that eventually.

PR #150, opened 2024-11-01
run-vmtest@v2 and build-selftests@v2

theihor:
Updates to various actions and scripts with breaking API changes. The goal of the changes is to make the actions usable by https://github.com/libbpf/libbpf/pull/862.

See commits for details.


patch-kernel/patch_kernel.sh
```diff
@@ -6,7 +6,7 @@ DIFF_DIR=$1
 for ext in diff patch; do
   if ls ${DIFF_DIR}/*.${ext} 1>/dev/null 2>&1; then
     for file in ${DIFF_DIR}/*.${ext}; do
-      if patch --dry-run -p1 -s < "${file}" 2>/dev/null; then
+      if patch --dry-run -N --silent -p1 -s < "${file}" 2>/dev/null; then
```
chantra, line 9: any reason for the -N here?

run-vmtest/run-bpf-selftests.sh
```diff
@@ -1,40 +1,26 @@
...
-# There is 2 ways to pass test names.
+# This script runs the tests within ${GITHUB_WORKSPACE}/selftests/bpf
+# If no specific test names are given, all test runners will be
+# executed. Otherwise, executed runners passed as parameters.
+# There are 2 ways to pass test names.
 # 1) command-line arguments to this script
-# 2) a comma-separated list of test names passed as `run_tests` boot parameters.
-# test names passed as any of those methods will be ran.
+# 2) a comma-separated list of test names passed as `run_tests` boot
+#    parameters.
```
chantra, line 18: this is probably more for historical reasons these days. Maybe we should support only one way. I believe the boot parameters approach is not used anymore since the move to vmtest. It was needed with the old image approach.

build-selftests/action.yml
```diff
@@ -1,34 +1,27 @@
...
   using: "composite"
   steps:
     - name: build selftests
       shell: bash
-      run: |
-        kbuild_output="$(realpath ${{ inputs.kbuild-output }})"
-        export LLVM_VERSION=${{ inputs.llvm-version }}
-        ${GITHUB_ACTION_PATH}/build_selftests.sh "${{ inputs.arch }}" "${{ inputs.kernel }}" "${{ inputs.toolchain }}" "${kbuild_output}"
       env:
-        MAX_MAKE_JOBS: ${{ inputs.max-make-jobs }}
```
chantra, line 34: this used to be used to control how many cpus were used when compiling. On bare metal, with 96 CPUs, if all containers are building at the same time, they will all use that many CPUs, bringing the machine to a crawl.

chantra (approved):

run-vmtest/run-bpf-selftests.sh
theihor: Yeah. It looks like boot parameters isn't used anymore. I'll clean this up.

patch-kernel/patch_kernel.sh
theihor: From man:

> -N  or  --forward
>           When a patch does not apply, patch usually checks if the patch
>           looks like it has been applied already by trying to reverse-
>           apply the first hunk.  The --forward option prevents that.
>           See also -R.

This is to avoid prints like this in the job logs:

> 1 out of 1 hunk ignored
> Reversed (or previously applied) patch detected!  Assume -R? [n] 
> Apply anyway? [n] 

IIRC `--silent` does not prevent this prompt.

build-selftests/action.yml
theihor: It's still used, it's just not an action input anymore. A caller can set it [like here](https://github.com/libbpf/libbpf/pull/862/files#diff-2ce184c8a6b8fb0bcdaaeb247ecef1e9d8e3badefc51505c2e5e731914cc93baR104).

The default is 4*nproc, [controlled here](https://github.com/libbpf/ci/blob/main/helpers.sh#L51-L57).

PR #156, opened 2024-11-26
setup-build-env: Install libzstd-dev

danobi:
sched-ext selftests link against zstd. It currently fails with the following linker error:

    rc/tools/testing/selftests/sched_ext/build/obj/sched_ext/test_example.o -lelf -lz -lpthread -lzstd
    /usr/bin/ld: cannot find -lzstd
    collect2: error: ld returned 1 exit status

Fix by installing zstd-dev package.

### DIFF
--- a/setup-build-env/action.yml
+++ b/setup-build-env/action.yml
@@ -24,7 +24,7 @@ runs:
       run: |
         echo "::group::Setup"
         sudo apt-get update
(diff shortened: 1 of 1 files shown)


danobi:
Tested in https://github.com/kernel-patches/bpf-rc/pull/4706

PR #160, opened 2024-11-28
build-linux: try container action

theihor:


theihor:
This would rely on docker-in-docker, which is not a nice thing to depend on. Nack.

PR #164, opened 2024-12-27
GCC BPF support

theihor:
Successor of https://github.com/libbpf/ci/pull/144


theihor:
@chantra this one makes GCC BPF specific changes. It should be reviewed/merged after #166

.github/workflows/kernel-build.yml
```diff
@@ -46,9 +46,13 @@ jobs:
...
     env:
         ARTIFACTS_ARCHIVE: "vmlinux-${{ inputs.arch }}-${{ inputs.toolchain_full }}.tar.zst"
+
+        BUILD_BPF_GCC: ${{ inputs.arch == 'x86_64' && 'true' || '' }}
+        BPF_GCC_INSTALL_DIR: ${{ github.workspace }}/bpf-gcc
+
         BPF_NEXT_BASE_BRANCH: 'master'
         BPF_NEXT_FETCH_DEPTH: 64 # A bit of history is needed to facilitate incremental builds
-        BUILD_SCHED_EXT_SELFTESTS: ${{ inputs.arch == 'x86_64' || inputs.arch == 'aarch64' && 'true' || '' }}
+        # BUILD_SCHED_EXT_SELFTESTS: ${{ inputs.arch == 'x86_64' || inputs.arch == 'aarch64' && 'true' || '' }}
```
chantra, line 55: is this related to this diff?

.github/workflows/kernel-build.yml
```diff
@@ -46,9 +46,13 @@ jobs:
     timeout-minutes: 100
     env:
         ARTIFACTS_ARCHIVE: "vmlinux-${{ inputs.arch }}-${{ inputs.toolchain_full }}.tar.zst"
+
+        BUILD_BPF_GCC: ${{ inputs.arch == 'x86_64' && 'true' || '' }}
```
chantra, line 50: isn't this going to build bpf-gcc for all combination of x86 builds?

shouldn't we instead define a "bpf-toolchain" that defaults to llvm (current behaviour), and can potentially be changd to gcc and use this build? Or is this handled transparently by the "test_progs-bpf_gcc" target?

chantra (commented):
LGTM, I am just not sure if the BPF-GCC is complementary to the current bpf build, e.g after this build, we have 2 distinctive bpf objects, 1 for the llvm build, and one for the gcc one?

.github/workflows/kernel-build.yml
theihor: No, I'll clean it up, thanks.

.github/workflows/kernel-build.yml
theihor: Right. This is going to build GCC for all x86 builds, but since it is cached, it's not an issue. When cached, it gets pulled in about 6 seconds: https://github.com/libbpf/ci/actions/runs/12777030966/job/35616953725

I thought about making a separate job (a dependency of kernel-test, when BUILD_BPF_GCC is set), since GCC doesn't really depend on anything withing our workflows. But I concluded it's won't make a difference, because the tests still depend on kernel to be built.

theihor:
@chantra please see an explanation re these two questions below

>  I am just not sure if the BPF-GCC is complementary to the current bpf build, e.g after this build, we have 2 distinctive bpf objects, 1 for the llvm build, and one for the gcc one?

> shouldn't we instead define a "bpf-toolchain" that defaults to llvm (current behaviour), and can potentially be changd to gcc and use this build? Or is this handled transparently by the "test_progs-bpf_gcc" target?

Yes, it's handled transparently in `selftests/bpf/Makefile`.

The "bpf-toolchain" doesn't really change when we set BPF_GCC. This is why I decided to not include the changes from #144 that add "bpf-toolchain" action arguments.

For example, let's say the CI toolchain-full is set to "llvm-17". So, the kernel is built with `LLVM=-17` ([see here](https://github.com/libbpf/ci/blob/main/build-linux/build.sh#L22)), and the test_progs binaries are built with llvm-17 (not [the default](https://github.com/libbpf/ci/blob/main/build-selftests/action.yml#L13)).

Now, on x86 `BUILD_BPF_GCC` is true, so BPF_GCC will be set to where freshly built gcc binaries are located. The `test_progs-bpf_gcc` binary is still built with clang 17 in this case ([make target just uses $(CC)](https://github.com/kernel-patches/bpf/blob/bpf-next_base/tools/testing/selftests/bpf/Makefile#L690)). What changes is:
1. `test_progs-bpf_gcc` is built ([by default it's not](https://github.com/kernel-patches/bpf/blob/bpf-next_base/tools/testing/selftests/bpf/Makefile#L745-L749))
2. The BPF objects *for test_progs-bpf_gcc* are built with provided GCC compiler. Later they are automagically included in the test runner binary as static arrays, IIRC.

So from the perspective of the CI, BPF_GCC is an *additional* test runner, built conditionally.

Now to the question wether we want to run this on all variants of the toolchain, I think yes. This way we are testing BPF emitted by GCC against kernels built by different toolchains.

chantra:
> 
> Now to the question wether we want to run this on all variants of the toolchain, I think yes. This way we are testing BPF emitted by GCC against kernels built by different toolchains.

Thanks, yes, make sense.

chantra (approved):

theihor:
vmtest CI passed: https://github.com/kernel-patches/vmtest/actions/runs/12814890364

Merging now.

PR #165, opened 2025-01-09
[draft] GCC BPF support

theihor:

PR #166, opened 2025-01-10
run-vmtest: introduce run-vmtest.env

theihor:
Update run-vmtest action to source $VMTEST_CONFIGS/run-vmtest.env when present. For detais read a comment in ci/vmtest/configs/run-vmtest.env

Add scripts local to run-vmtest that handle the merge of multiple allow/denylists for BPF selftests.

Passing @filename as allow/denylist to test runners, instead of inlining comma-separated test names.

Add test_progs-bpf_gcc as a possible test runner in run-vmtest/run-bpf-selftests.sh


theihor:
@chantra I got GCC BPF working at #165, and then decided to split it into a couple of PRs. Not all changes there are specific to the GCC support, like this one.

theihor:
vmtest CI passes: https://github.com/kernel-patches/vmtest/pull/326

.github/workflows/kernel-build.yml
```diff
@@ -87,23 +87,18 @@ jobs:
...
           llvm-version: ${{ inputs.llvm-version }}
           pahole: master
-      - name: Print toolchain version used
-        shell: bash
-        run: |
-          TOOLCHAIN=${{ inputs.toolchain }}
-          if [ $TOOLCHAIN = "llvm" ]; then
-            TOOLCHAIN="clang-${{ inputs.llvm-version }}"
-          fi
-          ${TOOLCHAIN} --version
```
chantra, line 106: is this intended to be part of this diff? We have used this in the past to troubleshoot CI breakages.

.github/workflows/kernel-build.yml
```diff
@@ -147,7 +143,7 @@ jobs:
         uses: ./tar-artifacts
         env:
           ARCHIVE_BPF_SELFTESTS: 'true'
-          ARCHIVE_MAKE_HELPERS: ${{ github.repository != 'kernel-patches/bpf' && 'true' || '' }}
```
chantra, line 150: I suppose this is also unrelated to this change also. Thinking about it, it probably does not make sense to special case making helpers or not in `tar-artifacts`.

run-vmtest/merge_test_lists.py
```diff
@@ -0,0 +1,41 @@
...
+        l1 = lines[i]
+        l2 = lines[j]
+        # for a pair of "foo_test" and "foo_test/subtest"
+        # remove a subtest from the list
+        if l1 == l2 or ('/' not in l1) and ('/' in l2) and l2.startswith(l1):
+            # print(f"removing '{l2}' because '{l1}' is in the list")
+            lines.remove(l2)
+        else:
+            i += 1
+            j += 1
```
chantra, line 33: LGTM, but this probably warrant some unittesting to validate it does the correct thing.

nit: `lines.remove` would probably cause a lot of trashing and we could use a generator here instrad of printing lines after `dedup_subtests` has ran, but for our use case, space/time complexity does not warrant the logic complexity it would introduce so we are better or leaving it as is.

run-vmtest/merge_test_lists.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/python3
```
chantra, line 1: maybe add a comment explaining what the script does? it does a bit more than just merging lists.

run-vmtest/prepare-bpf-selftests.sh
```diff
@@ -0,0 +1,30 @@
...
+    # first, append all the input lists into one
+    for file in "${files[@]}"; do
+        if [[ -f "$file" ]]; then
+            echo "cat $file >> $out"
+            cat "$file" >> "$out"
+        fi
+    done
+
+    # then merge the list of test names
+    cat "$out" | python3 "$(dirname $0)/merge_test_lists.py" > "$out"
```
chantra, line 20: I suppose this works because `cat` opens the file and get a `fd` before creating the redirectin of stdout to `$out`.

TBH, I think it is better to use a tmp file that gets deleted on exit just to avoid the potential race in what opens the file first.

here is what I mean:
```
echo a > /tmp/a
cat /tmp/a > /tmp/a
cat /tmp/a | wc -c
0
```

I am not even sure this works per:
```
$ echo a > /tmp/a
$ cat /tmp/a | wc -c
2
$ cat /tmp/a | cat  > /tmp/a
$ cat /tmp/a | wc -c
0
```

chantra (approved):
LGTM overall aside from the fact that I don't think this works correctly as is.

Is there a way to update https://github.com/libbpf/ci/blob/main/run-vmtest/action.yml or a README to document those environment variables needed by `run-vmtest.env`?


Also, I suppose this needs to be released in a version manner right? e.g landing this into `main` would break CI as https://github.com/kernel-patches/vmtest/blob/78dc9a42e4c3cdbad127814f91219ee20e40dbf2/.github/workflows/kernel-test.yml#L67 fir instance, is unaware of this yet.

run-vmtest/prepare-bpf-selftests.sh
chantra: maybe not given https://github.com/libbpf/ci/actions/runs/12716314984/job/35451077942#step:5:5276

but regardless.... it highlight the fact that this is prone to confusion :D

.github/workflows/kernel-build.yml
theihor: I try to keep `.github` and `ci` directories in sync between vmtest and libbpf/ci

This was part of vmtest, and at some point I removed it there, but not in libbpf/ci. This is kind of a debug print, I don't think we need it all the time. Also, the toolchain is printed as part of the job name.

.github/workflows/kernel-build.yml
theihor: This is also part of the vmtest sync. This had to be removed at some point, because of a change in kernel build scripts.

See https://github.com/kernel-patches/vmtest/commit/a0408d23ecbbbeee2c1d2b280023423c85c4e3f6

run-vmtest/merge_test_lists.py
theihor: Thanks. I will change to a generator.

I know it's not super important here, but it's we want good python here too.

run-vmtest/prepare-bpf-selftests.sh
theihor: Ok. We can just pass the file name as command line argument, and print the output. It'll be more clear.

theihor:
@chantra I added "unit tests" for run-vmtest, and rewrote the "merge lists" script. Please take a look when you get a chance.

run-vmtest/normalize_bpf_test_names.py
```diff
@@ -0,0 +1,66 @@
...
+        line = clean_line(line)
+        if len(line) == 0:
+            continue
+        input.append(line)
+    input.sort()
+    return input
+
+# Deduplicate subtests and yield the next unique test name
+def next_test(lines: list[str]):
+    prev = lines[0]
```
chantra, line 41: you need to ensure there is at least 1 line before calling this function. This was actually surfaced by the CI too.

chantra (approved):
Thanks, there is an issue in the rewritten script which was also surfaced by CI, but otherwise this looks go to go.

run-vmtest/normalize_bpf_test_names.py
theihor: woopsie

theihor:
Didn't answer these in the first pass through your comments. Answering below.

> Is there a way to update https://github.com/libbpf/ci/blob/main/run-vmtest/action.yml or a README to document those environment variables needed by run-vmtest.env?

It's tricky. `run-vmtest` action doesn't really care about the variables *used by* run-vmtest.env. It's simple "source", so whatever is in the file will be read.

Presumably, author of run-vmtest.env is also an author of a workflow calling `libbpf/ci/run-vmtest` action, and so they should know what variables to set.

On the other hand, there are `SELFTESTS_BPF_{ALLOW,DENY}LIST_FILES`, which is what run-vmtest.env must export (if it exists), because `prepare-bpf-selftests.sh` depends on this...

I think you're right, a note about this in the readme is warranted.

> Also, I suppose this needs to be released in a version manner right? e.g landing this into main would break CI as https://github.com/kernel-patches/vmtest/blob/78dc9a42e4c3cdbad127814f91219ee20e40dbf2/.github/workflows/kernel-test.yml#L67 fir instance, is unaware of this yet.

Yes. This is not an issue for a while now. BPF CI is pinned to `v2` currently, so a push to main won't break anything. I usually push to main, then check the update on vmtest, then update the current tag or push a new one (depending on wethere changes are significant).

theihor:
vmtest pipeline has passed: https://github.com/kernel-patches/vmtest/actions/runs/12776803904

Merging now.

PR #167, opened 2025-01-15
run-vmtest: fixes in run-bpf-selftests.sh

theihor:
test -f <file> && <command> is not exactly what we want: it returns non-0 exit code if file does not exists.

Replace them with an explicit if statement.

Example failure: https://github.com/libbpf/libbpf/actions/runs/12794559630/job/35669958975


theihor:
vmtest CI passed: https://github.com/kernel-patches/vmtest/actions/runs/12798491303

PR #168, opened 2025-02-03
Enable traffic monitoring to debug flaky selftests

ameryhung:
Traffic monitoring was add to the selftests/bpf[0] to debug some flaky test cases. Enable this in CI by building selftests with libpcap, installing libpcap in the test environment and finally turn it on when running the test.

[0] https://lore.kernel.org/bpf/20240815053254.470944-1-thinker.li@gmail.com/

### DIFF
--- a/run-vmtest/action.yml
+++ b/run-vmtest/action.yml
@@ -48,7 +48,7 @@ runs:
         foldable start install_qemu "Installing QEMU tools"
         # need gawk to support `--field-separator`
         sudo apt-get update && sudo apt-get install -y cpu-checker qemu-kvm qemu-utils qemu-system-x86 qemu-system-s390x qemu-system-arm qemu-guest-agent \
-          ethtool keyutils iptables \
+          ethtool keyutils iptables libpcap-dev \
           gawk
         foldable end install_qemu
     - name: Configure KVM group perms
--- a/run-vmtest/run-bpf-selftests.sh
+++ b/run-vmtest/run-bpf-selftests.sh
@@ -42,6 +42,7 @@ test_progs_helper() {
   args+=(${ALLOWLIST_FILE:+-a@$ALLOWLIST_FILE})
   args+=(${DENYLIST_FILE:+-d@$DENYLIST_FILE})
   args+=(-J "${json_file}")
+  args+=(-m '*')
 
   foldable start ${selftest} "Testing ${selftest}"
   echo "./${selftest}" "${args[@]}"
--- a/run-vmtest/run-bpf-selftests.sh
+++ b/run-vmtest/run-bpf-selftests.sh
@@ -25,11 +25,11 @@ OUTPUT_DIR=${OUTPUT_DIR:-/mnt/vmtest}
 
 test_progs_helper() {
   local selftest="test_progs${1}"
-  local args="$2"
+  local args=("$2")
 
-  args+=" ${TEST_PROGS_WATCHDOG_TIMEOUT:+-w$TEST_PROGS_WATCHDOG_TIMEOUT}"
-  args+=" ${ALLOWLIST_FILE:+-a@$ALLOWLIST_FILE}"
-  args+=" ${DENYLIST_FILE:+-d@$DENYLIST_FILE}"
+  if [ "${args[0]}" == "" ]; then
+	args=("${args[@]:1}")
+  fi
 
   json_file=${selftest/-/_}
   if [ "$2" == "-j" ]
@@ -38,12 +38,17 @@ test_progs_helper() {
   fi
   json_file="${OUTPUT_DIR}/${json_file}.json"
 
+  args+=(${TEST_PROGS_WATCHDOG_TIMEOUT:+-w$TEST_PROGS_WATCHDOG_TIMEOUT})
+  args+=(${ALLOWLIST_FILE:+-a@$ALLOWLIST_FILE})
+  args+=(${DENYLIST_FILE:+-d@$DENYLIST_FILE})
+  args+=(-J "${json_file}")
+
   foldable start ${selftest} "Testing ${selftest}"
-  echo "./${selftest} ${args} --json-summary \"${json_file}\""
+  echo "./${selftest}" "${args[@]}"
   # "&& true" does not change the return code (it is not executed
   # if the Python script fails), but it prevents exiting on a
   # failure due to the "set -e".
-  ./${selftest} ${args} --json-summary "${json_file}" && true
+  ./${selftest} "${args[@]}" && true
   echo "${selftest}:$?" >>"${STATUS_FILE}"
   foldable end ${selftest}
 }
--- a/.github/workflows/kernel-test.yml
+++ b/.github/workflows/kernel-test.yml
@@ -82,3 +82,9 @@ jobs:
           # Here we must use kbuild-output local to the repo, because
           # it was extracted from the artifacts.
           kbuild-output: ${{ env.REPO_ROOT }}/kbuild-output
+      - if: ${{ always() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: tmon-logs-${{ inputs.arch }}-${{ inputs.toolchain_full }}-${{ inputs.test }}
+          if-no-files-found: ignore
+          path: /tmp/tmon_pcap/*
--- a/setup-build-env/action.yml
+++ b/setup-build-env/action.yml
@@ -24,7 +24,7 @@ runs:
       run: |
         echo "::group::Setup"
         sudo apt-get update
-        sudo apt-get install -y cmake flex bison build-essential libssl-dev ncurses-dev xz-utils bc rsync libguestfs-tools qemu-kvm qemu-utils zstd libzstd-dev binutils-dev elfutils libcap-dev libelf-dev libdw-dev python3-docutils texinfo
+        sudo apt-get install -y cmake flex bison build-essential libssl-dev ncurses-dev xz-utils bc rsync libguestfs-tools qemu-kvm qemu-utils zstd libzstd-dev binutils-dev elfutils libcap-dev libelf-dev libdw-dev python3-docutils texinfo libpcap-dev
         echo "::endgroup::"
     - name: Install clang
       shell: bash


theihor:
@ameryhung If I understand correctly, the traffic monitoring logs are printed to files.

In github UI only stdout is going to be visible, the logs stored in files will be lost by default. So on CI you may want to either dump the logs to stdout, or collect them into a separate artifact.

ameryhung:
@theihor

I will change to upload the log file as well. Originally, I thought "-m" also prints to stdout. However, it does not include the payload. Having it should be more helpful for debugging.

theihor:
You can ignore GCC BPF tests failure. They are disabled at kernel-patches/bpf, I'll disable them here too shortly.

theihor:
I just noticed there is an older similar PR already: https://github.com/libbpf/ci/pull/130

There is a piece of yaml uploading logs as artifacts there. 

Re stdout vs artifacts, consider if the verbosity of the monitoring log. If it's lots of lines per test, then I'd go with the artifacts.

ameryhung:
I think uploading tmon logs as artifacts makes sense. I will update the PR. Thanks!

run-vmtest/action.yml
```diff
@@ -76,4 +80,9 @@ runs:
         VMLINUZ: ${{ inputs.vmlinuz || '' }}
       run: |
         ${GITHUB_ACTION_PATH}/run.sh
+    - uses: actions/upload-artifact@v4
+      with:
+        name: tmon-logs-${{ inputs.arch }}-${{ inputs.toolchain_full }}${{ inputs.kernel-test }}
+        if-no-files-found: error
+        path: "/tmp/tmon_pcap/*"
```
theihor, line 87: Managing artifacts is really a job-level task IMO. And we should avoid new action inputs if possible.

I think in this case, it makes sense to add an `always()` ([link](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always)) step in [kernel-test workflow](https://github.com/libbpf/ci/blob/main/.github/workflows/kernel-test.yml#L63) after run-vmtest call step. `toolchain_full` string is directly available there, since the build artifacts are identified by it. Also, with `always()` condition, logs will be uploaded on failure, which is what we want.

Later we may want to add other log files to the step (but don't worry about that in this PR).

theihor (changes requested):

run-vmtest/run-bpf-selftests.sh
```diff
@@ -30,6 +30,7 @@ test_progs_helper() {
   args+=" ${TEST_PROGS_WATCHDOG_TIMEOUT:+-w$TEST_PROGS_WATCHDOG_TIMEOUT}"
   args+=" ${ALLOWLIST_FILE:+-a@$ALLOWLIST_FILE}"
   args+=" ${DENYLIST_FILE:+-d@$DENYLIST_FILE}"
+  args+=" -m '*'"
```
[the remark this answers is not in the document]
ameryhung: This will enable traffic monitor only for the annotated tests. There are 11 such tests and the total log size for a single arch + compiler + test_progs flavor is 150KB. Is this acceptable?

run-vmtest/run-bpf-selftests.sh
```diff
@@ -43,7 +43,7 @@ test_progs_helper() {
   # "&& true" does not change the return code (it is not executed
   # if the Python script fails), but it prevents exiting on a
   # failure due to the "set -e".
-  ./${selftest} ${args} --json-summary "${json_file}" && true
+  ./${selftest} ${args} -m '*' --json-summary "${json_file}" && true
```
theihor, line 46: nit: adding `-m` to args was ok too

theihor (approved):
Great PR, thanks!

run-vmtest/run-bpf-selftests.sh
ameryhung: Do you mean what I had before (i.e., "args+=" -m '*'")?  That didn't work because the single quote inside the double quote was not evaluated and removed so test_progs would see '*' instead of *. I can replace the args with arrays if you prefer that way.

local args=("$2")

args+=("-m")
args+=("*")

./test_progs "${args[@]}"

theihor: > That didn't work because the single quote inside the double quote was not evaluated and removed so test_progs would see '' instead of *.

Hmm... that should work.

Consider this:
```bash
args+=" ${w:+-w$w}"
args+=" -m '*'"
echo ./foo "$args" bar
```

```shell
$ ./x.sh 
./foo   -m '*' bar
$ w=123 ./x.sh 
./foo  -w123 -m '*' bar
```

Anyways, this is a nit thing, don't spend too much time on it.

danobi:
IMHO uploading the artifact is not wrong and is generally useful. But it's probably a bit more useful to get the output from stdout/stderr. We already have infra in KPD that slurps failed selftest output and emails it out to contributors. Would be nice to leverage that.

If it's just a matter of improving the selftest infra in kernel tree then maybe @mlau2 can help.

theihor:
> IMHO uploading the artifact is not wrong and is generally useful. But it's probably a bit more useful to get the output from stdout/stderr. We already have infra in KPD that slurps failed selftest output and emails it out to contributors. Would be nice to leverage that.

@danobi dumping everything to stdout/stderr is ok up to a certain point. An extreme bad example is selftests/sched_ext which prints megabytes of log to stdout. This makes github ui to freeze, and is generally not convenient to work with.

The users of these logs should consider its verbosity, format and situations when it is actually needed. I assumed traffic monitoring might be pretty verbose, so uploading as separate artifacts made sense to me.

I just downloaded an artifact from a [successfull test run](https://github.com/libbpf/ci/actions/runs/13190882877), and it appears that the "logs" are actually some binary data (@ameryhung I don't really understand what it is, so if that's not right lmk):
```
$ cat packets-105-94-select_reuseport__sockhash_IPv4_TCP_LOOPBACK_test_err_sk_select-select_reuseport.log
�ò�oN�g��	XED��@@K���؅��� ����8��
�IH� oN�g��	<E(@@<�؅�����)P�h

$ xxd packets-105-94-select_reuseport__sockhash_IPv4_TCP_LOOPBACK_test_err_sk_select-select_reuseport.log
00000000: d4c3 b2a1 0200 0400 0000 0000 0000 0000  ................
00000010: 0004 0000 1401 0000 6f4e a667 8cd4 0900  ........oN.g....
00000020: 5800 0000 5800 0000 0800 0000 0000 0001  X...X...........
00000030: 0304 0006 0000 0000 0000 0000 4500 0044  ............E..D
00000040: f0c0 4000 4006 4bf1 7f00 0001 7f00 0001  ..@.@.K.........
00000050: 86c9 d885 dfec df20 0000 0000 a002 ffd7  ....... ........
00000060: fe38 0000 0204 ffd7 0402 080a 9249 48e6  .8...........IH.
00000070: 0000 0000 0103 0307 2000 0000 0000 0000  ........ .......
00000080: 6f4e a667 b0d4 0900 3c00 0000 3c00 0000  oN.g....<...<...
00000090: 0800 0000 0000 0001 0304 0006 0000 0000  ................
000000a0: 0000 0000 4500 0028 0000 4000 4006 3cce  ....E..(..@.@.<.
000000b0: 7f00 0001 7f00 0001 d885 86c9 0000 0000  ................
000000c0: dfec df29 5014 0000 9368 0000            ...)P....h..
```

If this is expected, then there is just no point in printing it to stdout: it's unreadable.

ameryhung:
@theihor The logs indeed are binary data and can be inspected using tcpdump. E.g., tcpdump -r <packet_log>

theihor:
> @theihor The logs indeed are binary data and can be inspected using tcpdump. E.g., tcpdump -X -r <packet_log> will show packets with their payloads

@ameryhung thank you, makes sense. Well it's 1.1Mb of the logs in total right now, and tcpdump-ing them is gonna be lots of lines. So let's use the artifacts.

We only need to make sure that relevant people are aware that this data is available.

iamkafai:
I just know I have a mlau2 account :)

In reply to @danobi about stdout/stderr, it does log to stdout/stderr (can't remember which one). I believe it was unnoticed because the test has passed. There is no details output by default for successful test.

Using "test_progs -v" should show it.  Like:

`
#> ./test_progs -v -m '*' -t tc_redirect/tc_redirect_dtime
test_inet_dtime:PASS:setns src 0 nsec
src     Out IPv4 172.16.1.100:48914 > 172.16.2.100:50001: TCP, length 60, SYN
src_fwd In  IPv4 172.16.1.100:48914 > 172.16.2.100:50001: TCP, length 60, SYN
dst     In  IPv4 172.16.1.100:48914 > 172.16.2.100:50001: TCP, length 60, SYN
dst_fwd Out IPv4 172.16.1.100:48914 > 172.16.2.100:50001: TCP, length 60, SYN
dst     Out IPv4 172.16.2.100:50001 > 172.16.1.100:48914: TCP, length 60, SYN, ACK
src     In  IPv4 172.16.2.100:50001 > 172.16.1.100:48914: TCP, length 60, SYN, ACK
dst     In  IPv4 172.16.1.100:48914 > 172.16.2.100:50001: TCP, length 52, ACK
dst_fwd In  IPv4 172.16.2.100:50001 > 172.16.1.100:48914: TCP, length 60, SYN, ACK
test_inet_dtime:PASS:connect_to_fd 0 nsec
src     Out IPv4 172.16.1.100:48914 > 172.16.2.100:50001: TCP, length 52, ACK
src_fwd Out IPv4 172.16.2.100:50001 > 172.16.1.100:48914: TCP, length 60, SYN, ACK
test_inet_dtime:PASS:accept 0 nsec
src_fwd In  IPv4 172.16.1.100:48914 > 172.16.2.100:50001: TCP, length 52, ACK
src     Out IPv4 172.16.1.100:48914 > 172.16.2.100:50001: TCP, length 68, ACK
test_inet_dtime:PASS:send to server 0 nsec
dst     In  IPv4 172.16.1.100:48914 > 172.16.2.100:50001: TCP, length 68, ACK
dst_fwd Out IPv4 172.16.1.100:48914 > 172.16.2.100:50001: TCP, length 52, ACK
__rcv_tstamp:PASS:recvmsg 0 nsec
`

danobi:
> @danobi dumping everything to stdout/stderr is ok up to a certain point. An extreme bad example is selftests/sched_ext which prints megabytes of log to stdout. This makes github ui to freeze, and is generally not convenient to work with.
>
> The users of these logs should consider its verbosity, format and situations when it is actually needed. I assumed traffic monitoring might be pretty verbose, so uploading as separate artifacts made sense to me.

Fair enough. In those cases the artifacts will be useful. And as far as I know, it's free on GHA.

danobi (approved):

PR #171, opened 2025-02-05
GCC BPF: factor out into a separate job

theihor:
In order to test selftests/bpf build with GCC BPF, the latest snapshot of GCC BPF compiler has to built from source.

We don't want this to be a dependency for all the test runners, so
introduce a separate gcc-bpf workflow that executes the following:
  * checkout Linux source tree and patch it (as in kernel-build)
  * download relevant artifacts with the KBUILD_OUTPUT objects (as in
    kernel-test)
  * build/download GCC BPF compiler
  * build selftests/bpf/test_progs-bpf_gcc runner, hence the test BPF
    objects built with GCC

The test_progs-bpf_gcc runner is not executed on BPF CI yet [1], so successful selftests build is a successful job run.

[1] https://lore.kernel.org/bpf/87bjw6qpje.fsf@oracle.com/

### DIFF
--- a/.github/workflows/gcc-bpf.yml
+++ b/.github/workflows/gcc-bpf.yml
@@ -1,0 +1,84 @@
+name: Testing GCC BPF compiler
+
+on:
+  workflow_call:
+    inputs:
+      runs_on:
+        required: true
+        type: string
+      arch:
+        required: true
+        type: string
+      llvm-version:
+        required: true
+        type: string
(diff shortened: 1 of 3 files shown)


theihor:
kernel-patches/vmtest pipeline has passed: https://github.com/kernel-patches/vmtest/actions/runs/13163688712/job/36739008180

PR #178, opened 2025-04-08
get-linux-source: always do git fetch

theihor:
Recently proof-of-work proxy was rolled out for git.kernel.org, which prevents wget of the bpf-next snapshots.

This lead to failures of get-linux-source action:
* https://github.com/kernel-patches/vmtest/actions/runs/14313967784/job/40115985369
* https://github.com/libbpf/libbpf/actions/runs/14316467048/job/40123652831

Simplify the download script to always run git init + git fetch sequence.

[1] https://social.kernel.org/notice/AsgziNL6zgmdbta3lY

### DIFF
--- a/.github/workflows/gcc-bpf.yml
+++ b/.github/workflows/gcc-bpf.yml
@@ -45,7 +45,7 @@ jobs:
 
       - if: ${{ inputs.download_sources }}
         name: Download bpf-next tree
-        uses: libbpf/ci/get-linux-source@v3
+        uses: ./get-linux-source
         with:
           dest: ${{ env.REPO_ROOT }}
           rev: ${{ env.BPF_NEXT_BASE_BRANCH }}
--- a/get-linux-source/checkout_latest_kernel.sh
+++ b/get-linux-source/checkout_latest_kernel.sh
@@ -1,69 +1,22 @@
 #!/bin/bash
 
-set -euo pipefail
+set -xeuo pipefail
 
-source $(cd $(dirname $0) && pwd)/../helpers.sh
-
-CWD=$(pwd)
-
-# 0 means just download a snapshot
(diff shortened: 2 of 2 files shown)


theihor:
Tesed on kernel-patches/vmtest https://github.com/kernel-patches/vmtest/actions/runs/14322033825

shunghsiyu:
@theihor the root cause here actually seems to be the good manner of setting the user agent to `BPFCIBot/1.0 (bpf@vger.kernel.org)`. If that was dropped `wget` actually downloads the file just fine.

I'll ask in [the announcement](https://social.kernel.org/notice/AsgziNL6zgmdbta3lY) whether that was intended.

shunghsiyu:
The [suggestion](https://social.kernel.org/notice/AtCyOpgtzYUaqxbkoa) is to use `BPFCI/1.0` as the user agent instead. Testing that out locally it seems to work just fine.

theihor:
@shunghsiyu Hi, thank you for the tip.
In practice downloading a snapshot vs fetching doesn't make much of a difference in terms of download time. So switching back to snapshots only makes sense if kernel.org collects statistics on external hits. Do you know if it does?

anakryiko:
@theihor Konstantin (kernel.org admin at that time) did have a preference for snapshot downloading, it's probably causing much less load on their servers. I forwarded you some emails we exchanged a while ago for some context. So yes, snapshots are preferred.

theihor:
> @theihor Konstantin (kernel.org admin at that time) did have a preference for snapshot downloading, it's probably causing much less load on their servers. I forwarded you some emails we exchanged a while ago for some context. So yes, snapshots are preferred.

Thanks, I wasn't aware. I find it strange to trigger proof-of-work challenge on "bot" string in the user-agent though.

PR #179, opened 2025-04-17
setup-build-env: Fix cross compilation toolchain

theihor:


theihor:
Apparently, downgrade of linux-libc-dev:amd64 helps: https://github.com/kernel-patches/vmtest/actions/runs/14520463948/job/40756555954

Merging this in.

PR #180, opened 2025-04-22
setup-build-env: do not downgrade linux-libc-dev

theihor:
The downgrade helped mitigate package conflicts between linux-libc-dev:amd64 and s390x packages.

Revert the downgrade command.


theihor:
`setup-build-env` step was successful on s390x, merging.
https://github.com/kernel-patches/vmtest/actions/runs/14605970922/job/40974951713

PR #184, opened 2025-05-15
run-vmtest: exit gracefully from print_test_summary.py

theihor:
When selftests fail because of a watchdog timeout or segfault or anything similar, a JSON summary might be empty or absent.

Currently this leads to print_test_summary.py to verbosly fail with an exception, which is confusing when looking at the job log.

Handle such situations gracefully in the script.

### DIFF
--- a/run-vmtest/print_test_summary.py
+++ b/run-vmtest/print_test_summary.py
@@ -9,7 +9,8 @@
 
 import argparse
 import json
-
+import os
+import sys
 
 def parse_args():
     parser = argparse.ArgumentParser()
@@ -75,14 +76,28 @@ def test_error_console_log(test_error: str, test_message: str) -> str:
     else:
         return error_msg
 
+def error_die(msg: str):
+    print("print_test_summary.py: {}".format(msg), file=sys.stderr)
+    exit(0)
 
 if __name__ == "__main__":
     args = parse_args()
     step_open_mode = "a" if args.append else "w"
+    str_summary = None
     json_summary = None
 
+    if not os.path.exists(args.json_summary):
+        error_die("Could not find {}".format(args.json_summary))
+    elif os.stat(args.json_summary).st_size == 0:
+        error_die("{} is empty".format(args.json_summary))
+
     with open(args.json_summary, "r") as f:
-        json_summary = json.load(f)
+        str_summary = f.read()
+
+    try:
+        json_summary = json.loads(str_summary)
+    except json.JSONDecodeError:
+        error_die("{} is not a valid JSON\n{}".format(args.json_summary, str_summary))
 
     with open(args.step_summary, step_open_mode) as f:
         log_gh_summary(f, "# Tests summary")


run-vmtest/print_test_summary.py
```diff
@@ -75,31 +75,41 @@ def test_error_console_log(test_error: str, test_message: str) -> str:
...
-            for subtest in test["subtests"]:
-                if subtest["failed"]:
-                    subtest_log = f"#{test_number}/{subtest['number']} {test_name}/{subtest['name']}"
-                    log_gh_summary(f, subtest_log)
-                    log_console(test_error_console_log(subtest_log, subtest["message"]))
+    try:
+        with open(args.json_summary, "r") as f:
+            json_summary = json.load(f)
+    except:
+        report_error("Could not load JSON summary file {}".format(args.json_summary))
```
eddyz87, line 90: Is there a way to get actual parsing error from the exception?

theihor: It's usually `JSONDecodeError` that we could catch, but then any other failure would result in the traceback splat.

I think it's not important to distinguish between errors within this script. It only fails if something went wrong with the tests.

eddyz87: Is there still a way to infer why operation failed? As far as I understand, this completely hides exception.

theihor: > Is there still a way to infer why operation failed? 

If you mean the script itself, then yeah, we could handle different exceptions differently. But I don't see much benefit to that. The reason for this change is to reduce noise. "File not found" or "summary file is empty" error  doesn't help in the context of CI.

If you mean "can we print a useful tests summary even if json was not generated?", that's a different problem. I think that would require capturing stdout/stderr of the vmtest program.

eddyz87: No, I mean, suppose something is wrong with json printing or the file is missing, would there be a log anywhere saying that json is malformed or absent, or "could not load" is all you'd get?

theihor: > No, I mean, suppose something is wrong with json printing or the file is missing, would there be a log anywhere saying that json is malformed or absent, or "could not load" is all you'd get?

The way the change is written you'll only see the "could not load" message, yes. I want to get rid of useless traceback we often see in the job log.

Are you worried that we may miss a malformed json produced by a test runner? Has that ever happened?

eddyz87: > Are you worried that we may miss a malformed json produced by a test runner? Has that ever happened?

No, just would be harder to figure out what happened. Point being, just add `str(e)` to the error string?

theihor: I changed to handling "expected" errors: no file, empty file, json decode error. Anything else should produce a stacktrace. Thanks for taking a look.

eddyz87 (approved):

PR #195, opened 2025-09-25
rootfs: Add libpcap to rootfs.

higuoxing:
Commit: f52403b ("selftests/bpf: Add traffic monitor functions.") introduces libpcap to bpf selftests.  It would be good to add libpcap to rootfs.


theihor:
Hi @higuoxing 

Do you use the `rootfs/` scripts yourself?

The rootfs scripts haven't been used by BPF and libbpf CI for a while now. We switched to [vmtest](https://github.com/danobi/vmtest) which doesn't require explicit rootfs.

PR #224, opened 2026-06-05
ansible: drop Amazon Linux 2 / RedHat support after Ubuntu migration

theihor:
The AWS bare-metal runner fleet has been migrated from Amazon Linux 2 (RedHat family) to Ubuntu 24.04, and AL2 is decommissioned. Every host is now Debian family (Ubuntu EC2 and s390x LinuxONE both use the apt path), so the RedHat/Amazon-conditional code is unreachable. Remove it and collapse the now single-OS base role.

- base: delete the RedHat task/vars files and the ansible_os_family include dispatch; fold the package list (incl. docker.io) into defaults and inline the Debian tasks into tasks/main.yml.
- runner: drop the "Install docker pip on Amazon Linux" task and the tautological `ansible_os_family == 'Debian'` guard; simplify the EC2 instance-id block to `ansible_system_vendor == 'Amazon EC2'`.
- docs: base README is now Debian-only; ansible README install uses apt.

amazon.aws (ec2_metadata_facts) and community.docker (docker_login) are still required and kept. The qemu-user-static role is left unchanged: it remains live on the s390x hosts (registers binfmt_misc handlers).


theihor:
This is already deployed, runners are good. Merging.

PR #227, opened 2026-07-29
build-bpf-gcc: add gcc-version input to select the GCC major version

vineetgarc:
The GCC major version was hardcoded to 15 in latest-snapshots.sh. Add a 'gcc-version' action input (default "15", so existing callers are unchanged) that is passed through as $GCC_VERSION and used to pick the LATEST-<version> snapshot directory and tarball. This lets callers (e.g. theihor/gcc-bpf) build GCC 16 for BPF CI toolchain experiments without forking the action.


theihor (approved):
LGTM

PR #229, opened 2026-08-10
run-vmtest: fail selftest runs that produce kernel splats

mykyta5:
A non-fatal splat (WARN, KASAN report, lockdep, hung task) leaves the VM running, so the selftest binary still exits 0 and CI goes green on a kernel that just hit a real bug. We have no signal for these today.

Scan dmesg after the run and add a kernel_splats row to the existing exitstatus aggregator. A scan that cannot run fails the row rather than reporting clean, otherwise a typo in a pattern file would quietly turn the check off. splat-allowlist-file and skip-kernel-splat-check provide an escape hatch for known-benign matches.


run-vmtest/check-kernel-splats.sh
```diff
@@ -0,0 +1,58 @@
...
+#
+# The lockups get their own branch: kernel/watchdog.c sets pr_fmt, so the line
+# starts with `watchdog: `, not with `BUG:`.
+splat='^(\[[^]]*\] *)*((BUG|WARNING|UBSAN|Oops)[: ]|kernel BUG at'
+splat+='|watchdog: .*(soft lockup|hard LOCKUP)'
+splat+='|(rcu: )?INFO: (task .* blocked for more than|[_a-z]+ (self-)?detected stall))'
+
+# The selftest config leaves unprivileged BPF on, so the Spectre v2 code warns
+# about it on every boot. It states a config choice, not a kernel bug.
+benign='WARNING: Unprivileged eBPF is enabled'
```
theihor: I'm thinking this should be configurable through ci/vmtest/configs, like denylists and co. This will make it easier to control the splat matching - with only a change in [vmtest](https://github.com/kernel-patches/vmtest/tree/master/ci/vmtest/configs).

.github/workflows/kernel-test.yml
```diff
@@ -96,3 +96,12 @@ jobs:
...
           path: /tmp/tmon_pcap/*
+
+      # Written by check-kernel-splats.sh inside the VM, into the bind-mounted
+      # workspace. Needed to triage a splat, and to audit the scan itself.
+      - if: ${{ always() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: kernel-log-${{ inputs.arch }}-${{ inputs.toolchain_full }}-${{ inputs.test }}
+          if-no-files-found: ignore
+          path: dmesg.txt
```
theihor, line 107: This will need to be copied over to the [vmtest](https://github.com/kernel-patches/vmtest/blob/master/.github/workflows/kernel-test.yml) repo.

theihor (changes requested):
I created a vmtest PR for testing, and apparently there is pre-existing arm64 kasan splat: https://github.com/kernel-patches/vmtest/actions/runs/31545541634/job/93958798153?pr=508

That's good: the splat detection is working. But it also means we can't merge this before that splat is fixed. Do you want to look into it?

Alternatively, we can land this before the upstream fixes as non-blocking by reporting a warning or printing to summary. It will be mostly ignored then though.

.github/workflows/kernel-test.yml
mykyta5: I'm looking into that pre-existing splat.
moving regexes to separate files - makes sense.
Thanks for taking alook.

theihor (approved):

PR #233, opened 2026-08-13
re-run denylisted tests for test_progs-bpf_gcc

vineetgarc:
GCC denylist is static and tests listed there are just never run. This means it can potentially miss out improvements due to compiler/selftest updates/improvements.

So ran an additional second pass over just denylist as allowlist. This run is purely informational: its exit code is deliberately not appended to the exit status file, so tests that are still failing, which is the expected outcome for most of them, cannot fail the job.


etsal (approved):
Some duplication there with test_progs_helper, but we can't help it.

